PR #71, opened 2015-05-06
Support for resuming staging transfers

natefoo:
Only implemented in the curl transport though.

Client changes in Galaxy in galaxyproject/galaxy#227

### DIFF
--- a/pulsar/client/transport/curl.py
+++ b/pulsar/client/transport/curl.py
@@ -66,10 +66,18 @@ def post_file(url, path):
 
 
 def get_file(url, path):
+    if path and os.path.exists(path):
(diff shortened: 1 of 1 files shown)


jmchilton:
Any particular reason to add an option and not just have the curl down-loader just do this by default?

natefoo:
I... no... err... I guess I just thought there might be some people who didn't want it enabled by... yeah, probably not.

Well, I know more about how Pulsar works now, anyway. :sob:

natefoo:
@jmchilton Okay, got rid of the configurability.

jmchilton:
Awesome - thanks! And sorry you had to learn so much Pulsar.

PR #76, opened 2015-08-26
Add a job manager for XSEDE jobs

natefoo:
XSEDE would like us to run their job/user accounting script after submission.

TODO: A generalized callback framework for executing things at various points in the job lifecycle.


jmchilton:
Sorry for the delay, I need to figure out what is broken in Travis I guess at some point - but it doesn't seem related to this at all.

PR #78, opened 2015-09-08
Added the slash for the self-closed plugin pulsar tag

remimarenco:
Added the slash for the self-closed plugin pulsar tag in the file `job_conf_sample_remote_cluster.xml` located in `docs/files/`, otherwise we get the Exception:

```
Exception: Problem parsing the XML in file /Users/marenc01/Dev/galaxy/config/job_conf.xml, please correct the indicated portion of the file and restart Galaxy.mismatched tag: line 6, column 6
```

When running Galaxy.


bgruening:
:+1:

PR #79, opened 2015-09-08
Updated the job_conf_sample_remote_cluster.xml for the right default …

remimarenco:
…`destinations` id

In the `job_conf_sample_remote_cluster.xml`, the default destinations attribute was set on `drmaa` instead of `local_cluster`.

This was producing the following Exception:

```
Exception: Problem parsing the XML in file /Users/marenc01/Dev/galaxy/config/job_conf.xml, please correct the indicated portion of the file and restart Galaxy.<destinations> default attribute 'drmaa' does not match a defined id or tag in a child element
```

When launching Galaxy.

### DIFF
--- a/docs/files/job_conf_sample_remote_cluster.xml
+++ b/docs/files/job_conf_sample_remote_cluster.xml
@@ -7,7 +7,7 @@
     <handlers>
         <handler id="main"/>
     </handlers>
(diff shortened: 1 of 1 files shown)


jmchilton:
Thanks @remimarenco.

Issue #77, opened 2015-09-02
MQ messages are never received


natefoo:
In our setup, connections to the RabbitMQ server seem to be less than 100% reliable. I am not sure why this is, but the retry options seem to have addressed most problems. However, it turns out to be the case that when publishing a message, unless there is a connection problem, AMQP itself does not guarantee that your message has been received *by the AMQP server* after the client's publish call completes. This explains why sometimes, Pulsar's status updates seem to be going in to a void.

Transactions are an AMQP standard and would potentially solve this, but not terribly popular and there was discussion of removing them from RabbitMQ entirely. The RabbitMQ preferred solution for guaranteeing that a message has been received is using [Publisher Confirms](http://www.rabbitmq.com/blog/2011/02/10/introducing-publisher-confirms/). However, in testing this feature, I seem to be able to enable it, but I can't find a way to actually handle the "not acknowledged" messages that I think are being generated (unless they are not - debugging this is infuriatingly complicated)...

Using Pulsar's `PulsarExchange` class, one can publish a message like so:

```
exchange.publish('queue_amqp_test', {'msg':'foo'}, connection_kwargs={'transport_options':{'confirm_publish':True}})
```

Do that in a loop and the rate of delivery will be much slower than without `confirm_publish` set, which I believe suggests that the confirms are being sent. However, when the [queue is killed to generate the nacks](http://rabbitmq.1065348.n5.nabble.com/Testing-lightweight-publisher-confirms-td834.html), the messages just keep going into the void, no exceptions are raised, and, if an `on_return` callback is set on the `Producer` object [as the fine manual suggests might kinda be relevant except maybe not since it's maybe for the *mandatory* or *immediate* arguments which it then later says are not supported anyway](http://kombu.readthedocs.org/en/latest/reference/kombu.html#message-producer), that callback is never fired.

Because of this, and the fact that confirms are a RabbitMQ-specific feature anyway, I think we may need to solve this problem application-side by having Galaxy send messages back to Pulsar acknowledging receipt, like so:

```
+-----------------+
| Pulsar          |
+-----------------+
 |   |   ^   |   ^       
[1] [2]  |  [2]  |       
 |   |  [3]  |  [4] 
 v   v   |   v   |
+-----------------+
| Galaxy          |
+-----------------+

[1]: status update sent but not received
[2]: status update resent after some wait, and received
[3]: acknowledgement of status update sent but not received
[4]: acknowledgement of status update sent and received
```

[...]

PR #81, opened 2015-09-08
Ensure AMQP consumers receive published messages

natefoo:
See #77 for rationale.


natefoo:
Galaxy side of the work is in galaxyproject/galaxy#704

PR #84, opened 2015-09-15
Fix config.py on Windows.

ssorgatem:
config.py was failing on Windows due to missing attributes in the PlatformArgumentParser. As these are skipped for Windows, I just added a check so it doesn't try to access them on Windows.

With these little changes, pulsar-config worked for me on Windows. 

This fix is the least intrusive IMO, but I think a better solution would be to use hasattr, but I don't know if that would break any assumption by the rest of the code.


jmchilton:
This looks great to me, thanks @ssorgatem!!

PR #95, opened 2016-04-05
Fixes for things encountered when running jobs on Jetstream

natefoo:
Not sure why embedded Pulsar encounters the first issue (attempting to create the "remote" job dir after having already created it to stage inputs) when standalone Pulsar does not, but this fix should be harmless.

And update cluster slot handling as in galaxyproject/galaxy#1347.

Might not want to merge this yet as more may be discovered.


natefoo:
This is all I'm gonna have. Merge at will.

natefoo:
@jmchilton should I do a full import of the jobs stuff from Galaxy or is it okay to just copy over the cluster slots script?

PR #97, opened 2016-04-06
Use a global DRMAA session

natefoo:
Only one active DRMAA session is possible per process. This allows multiple Galaxy runner plugins to be active in the same handler if Pulsar is being used as an embedded runner (they all have to target the same DRMAA library, however). Also add locking to calls that modify the number of active jobs, as per the Python-drmaa documentation.

See also galaxyproject/galaxy#2102

### DIFF
--- a/pulsar/managers/util/drmaa/__init__.py
+++ b/pulsar/managers/util/drmaa/__init__.py
@@ -49,7 +49,6 @@ class DrmaaSession(object):
                 DrmaaSession.session.initialize()
             DrmaaSession.session_count += 1
 
-
     @contextlib.contextmanager
     def _session_lock(self):
         with DrmaaSession.session_lock:
--- a/pulsar/managers/util/drmaa/__init__.py
+++ b/pulsar/managers/util/drmaa/__init__.py
@@ -1,3 +1,6 @@
+import threading
+import logging
+
 try:
     from drmaa import Session, JobControlAction
 except OSError as e:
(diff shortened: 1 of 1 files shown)


natefoo:
@jmchilton if the `DrmaaSession` is set up in some different way for the tests and I've broken it (perhaps by changing the the session constructor to be passed in uncalled?) I can't find where. Everything works when you actually run Galaxy (I didn't test a standalone Pulsar).

pulsar/managers/util/drmaa/__init__.py
```diff
@@ -35,14 +36,23 @@ class DrmaaSession(object):
...
     def __init__(self, session_constructor, **kwds):
-        with DrmaaSession.session_lock:
+        with self._session_lock():
             if DrmaaSession.session is None:
+                if DrmaaSession.session_count != 0:
+                    log.warn("DrmaaSession.session is None but session_count is non-zero - logic error occurred.")
                 log.debug("Initializing DRMAA session from thread %s", threading.current_thread().name)
                 DrmaaSession.session = session_constructor()
                 DrmaaSession.session.initialize()
+                DrmaaSession.session_count += 1
```
natefoo: This needs to move up one level right? Otherwise it will only ever be 1.

jmchilton: Yes :frowning:

pulsar/managers/util/drmaa/__init__.py
```diff
@@ -41,13 +41,14 @@ class DrmaaSession(object):
 
     def __init__(self, session_constructor, **kwds):
         with self._session_lock():
+            DrmaaSession.session_count += 1
```
natefoo: Heh... but now it will trigger the warning on line 47. This should come after the if block starting on line 45.

jmchilton: Well clearly this coding thing isn't working out for me today - this would be a good day for that Pittsburgh trip.

PR #104, opened 2016-04-15
Allow per-manager job directories.

jmchilton:


mvdbeek:
Just when I needed it, i'll give it a try.

mvdbeek:
I guess this would work, but I'm having trouble setting up named managers in #105 
Also, could you add an example to app.yml.sample ?

jmchilton:
@mvdbeek Thanks for the suggestion - rather than document right in ``app.yml.sample`` I added documentation for this feature in ``docs/job_managers.rst`` and linked to this general manager configuration information on RTD from ``app.yml.sample``.

PR #126, opened 2016-12-12
Conda dependency resolution

jmchilton:


bgruening:
:tada: 
John do we need to add Windows as well to: https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/tools/deps/conda_util.py#L35

Otherwise there will be no conda available for Windows isn't it?

Plans to also support BioContainers?

jmchilton:
Pre-built BioContainers might already work correctly - I guess some hacking would be needed to get the dynamic ones working.

Windows support for dependency resolution would be a welcome contribution but is outside the scope of what I intend to work on.

natefoo:
I give this PR 5 snakes up: :snake::snake::snake::snake::snake:

PR #130, opened 2017-05-12
Update job_managers.rst

mapa17:
Missing quotes causes yml validation error


jmchilton:
Awesome - thanks for the fix!

PR #136, opened 2017-09-15
Support the remote_transfer file action without setting the jobs_directory destination param

natefoo:
Fixes the `job directory required for action FileAction[...` error. All of the necessary params are available from the remote server in the job config, and setting the `jobs_directory` param disables querying remote params, so either way we'll know what the job directory is.


jmchilton:
Rebase?

PR #139, opened 2017-09-19
Additional testing fixes

natefoo:


jmchilton:
For the record it looks like the root problem is that Travis switched the testing VM away from ``precise``. There are still some sudo issues that are new due to the switch but they are certainly unrelated to your PRs.

natefoo:
Yeah, the default is now `trusty`.

PR #141, opened 2017-09-20
Precreate Galaxy tool outputs on the remote before executing.

natefoo:
When using remote staging with retries, if a job fails and never writes to its expected output(s), Pulsar will retry output collection on outputs that will never exist, and fail the job once it gives up. This change causes them to be created before the job is executed, mirroring the behavior in Galaxy.

### DIFF
--- a/pulsar/client/client.py
+++ b/pulsar/client/client.py
@@ -120,6 +120,9 @@ class JobClient(BaseJobClient):
             launch_params['env'] = json_dumps(env)
         if remote_staging:
             launch_params['remote_staging'] = json_dumps(remote_staging)
+        if job_config and 'touch_outputs' in job_config:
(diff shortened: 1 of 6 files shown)


natefoo:
I'm not really excited about `submit_extras` but couldn't come up with a better solution.

jmchilton:
Wait... yeah is submit extras the wrong place to do this. Hmm...

natefoo:
It's not really a setup param and it's not really a submit param. It's in the `job_config`, but the `job_config` is not passed to the `submit_job` manager endpoint unless using AMQP. And I didn't want to pass it when using REST lest it break things in other places.

natefoo:
Anyway, thanks for the merge.

PR #144, opened 2017-09-21
Release 0.8.0

natefoo:
My first time releasing so I figured I'd PR it for review.


jmchilton:
I should have caught this sooner - but this adds new features - it should be 0.8.0 right? If that isn't easy to fix though we can merge as is.

natefoo:
Those new features are pretty tiny and mostly have to do with remote metadata which no one uses. But yeah, it's no problem for me to bump to 0.8.0.

PR #147, opened 2017-09-28
Strip quotes from around filenames when matching

blankenberg:
Fixes issues caused by a hanging `'` when quotes were used around a `$__tool_directory__` file.


jmchilton:
Thanks for the fix Dan. This breaks the test under Python 3 since map returns an iterator and not a list and I'm not sure I love doing it after the fact like that. I opened #148 that modifies the regex patterns to catch these at that level instead and includes a test for the functionality. Hope that is okay - sorry that Pulsar wasn't catching those files.

blankenberg:
Cool, thanks for the fix :)

Issue #151, opened 2017-12-06
Send stdout/stderr as job files when large


natefoo:
I've had a few cases where the AMQP message generated at job completion is enormous (~6GB) due to excessive program output. We should transfer these as job files when they are over a configurable limit (or maybe always).

PR #157, opened 2018-02-14
Add configuration parameter to limit stream size read from disk.

jmchilton:
Set with maximum_stream_size in app.yml.

Default to 1MB.

Fixes underlying problem in #151 I think, though doesn't actually do what was requested. I prefer this - is that okay @natefoo?

### DIFF
--- a/pulsar/client/test/check.py
+++ b/pulsar/client/test/check.py
@@ -33,16 +33,19 @@ from pulsar.client import (
 from .test_common import write_config
 
 TEST_SCRIPT = b"""# -*- coding: utf-8 -*-
+from __future__ import print_function
+
 import sys
+
 from os import getenv
-from os import makedirs
 from os import listdir
-from os.path import exists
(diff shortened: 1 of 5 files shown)


natefoo:
It'd be nice if we could avoid losing the stdout/stderr as the error message you'd use to determine why a job failed is likely at the end. But this will at least avoid the biggest problem, which is RabbitMQ crashing on huge messages.

jmchilton:
> It'd be nice if we could avoid losing the stdout/stderr as the error message you'd use to determine why a job failed is likely at the end. 

Agreed.

And another next step that would be nice from an API perspective is to avoid the concatenation that Galaxy does. I'd rather have structured information - like ``{..., "stdout_contents": "...", "stdout_truncated": true, "stdout_contents_end": "..."}`` in this API and then let Galaxy mangle the data I think.

natefoo:
What's the concatenation?

PR #164, opened 2018-03-05
Rework preprocessing job state.

### LINKED ISSUE #161: Pulsar loses jobs if restarted during job setup

jmchilton:
- Instead of persisting job metadata for preprocessing - write out metadata for preprocessing and launching.
- Track list of jobs currently being "preprocessed" - not just launched.
- Use the list to attempt to "recover" preprocessing jobs after restart.
- Skip preprocessing complexity all together (extra metadata in the job for instance and this new list) if not preprocessing steps are needed.
- Test case for new behavior.

Should fix #161.

### DIFF
--- a/pulsar/manager_endpoint_util.py
+++ b/pulsar/manager_endpoint_util.py
@@ -9,7 +9,6 @@ from pulsar import __version__ as pulsar_version
 from pulsar.client.setup_handler import build_job_config
 from pulsar.managers import status
 from pulsar.managers import PULSAR_UNKNOWN_RETURN_CODE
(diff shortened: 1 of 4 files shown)


jmchilton:
Ping @natefoo - so this isn't backward compatible yet - jobs that are running with remote post-processing instructions won't find them in the right place after a restart and upgrade of Pulsar. Do I need to fix that or are there ever upgrade times with no running jobs?

natefoo:
I can upgrade w/ no running jobs. Thanks for this!

PR #171, opened 2019-03-20
Update job_conf_sample_mq.xml

gmauro:
fix parameter label


jmchilton:
Awesome, thanks so much for the fix!

PR #181, opened 2019-05-14
Use wildcard char for pulsar version.

VJalili:


jmchilton:
Works for me, didn't know ADD could do that - thanks so much!

PR #182, opened 2019-05-15
Remove previous wheels before creating a new one.

VJalili:
Since there can be wheels belonging to different version of pulsar already existing in that folder, and since that can cause issues for building docker image as it uses `*`, hence it would be better to empty the folder of any previous wheels first.

### DIFF
--- a/docker/coexecutor/Makefile
+++ b/docker/coexecutor/Makefile
@@ -1,6 +1,7 @@
 
 
 docker-image:
+	find . -regex './pulsar_app-.*.whl' -delete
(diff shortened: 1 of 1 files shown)


jmchilton:
Encountered the same thing, but fixed it by including deleting the wheel in the ``make dist`` rule -> ``make clean``.

VJalili:
@jmchilton both methods seem reasonable to me; we can implement whichever you prefer.

bgruening:
@VJalili `make clean` seems to me better.

PR #185, opened 2019-05-16
Fix 'amqp_ack_republish_time' in sample

dannon:
(from 'amqp_republish_time' which is unused)

xref https://github.com/galaxyproject/galaxy/pull/7983


jmchilton:
Thanks @dannon !

PR #166, opened 2018-06-30
Add a Dockerfile for Pulsar with CVMFS

nuwang:
xref: https://github.com/galaxyproject/galaxy/issues/6426

PR #195, opened 2019-06-18
Attempt to merge Pulsar Docker images (1 / ...)

jmchilton:
I was working on getting the Kubernetes support to just work with a single container and container resolution (after being prodded to do this by @nuwang). I got it working with Conda environments inside the container - but obviously being able to leverage Conda from CVMFS would be beneficial as would being able to leverage reference data from my Kubernetes work. So I'm trying to merge the differences between our containers down. If they need to be different - hopefully one can just depend on the other or they can both depend on some common base container. To make sharing files easier I'm trying to move scripts and data used into ``pulsar.scripts`` - so installing Pulsar results in this stuff being available and being self contained (no need to ensure things are below the Dockerfile base path, etc..).


jmchilton:
Merging because there is an important fix in here - the cvmfs Docker image still doesn't build because of https://github.com/galaxyproject/pulsar/pull/166#discussion_r295379856

Issue #194, opened 2019-06-17
assign_ids: uuid produces an IOError: [Errno 2] No such file or directory


gmauro:
I have set assign_ids: uuid in the app.yml file and the result has been an IOError: [Errno 2] No such file or directory:

```
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: 2019-06-17 14:36:41,898 ERROR [pulsar.messaging.bind_amqp][consume-setup-pyamqp://galaxy_freiburg:********@proxy.internal.galaxyproject.eu:5671//pulsar/galaxy_freiburg?ssl=1] Failed to setup job 161 obtained via message queue.
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: Traceback (most recent call last):
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: File "/opt/pulsar/venv/lib/python2.7/site-packages/pulsar/messaging/bind_amqp.py", line 117, in __process_setup_message
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: manager_endpoint_util.submit_job(manager, body)
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: File "/opt/pulsar/venv/lib/python2.7/site-packages/pulsar/manager_endpoint_util.py", line 102, in submit_job
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: manager.preprocess_and_launch(job_id, launch_config)
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: File "/opt/pulsar/venv/lib/python2.7/site-packages/pulsar/managers/stateful.py", line 85, in preprocess_and_launch
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: self._persist_launch_config(job_id, launch_config)
Jun 17 14:36:41 vgcn-central-manager.usegalaxy.eu.novalocal pulsar[333945]: File "/opt/pulsar/venv/lib/python2.7/site-packages/pulsar/managers/stateful.py", line 76, in _persist_launch_config
[...]

PR #196, opened 2019-06-19
 enable uuid job_id with Pulsar over Message Queue 

gmauro:
This commit creates a random id in this way: galaxy_job_id-uuid4, e.g.:
185-c83c0087b9cc4adf982b4697c0da803d
changes paths in the command_line and remote_staging accordingly and then use the galaxy_job_id only to publish amqp update messages.

This solves #194

### DIFF
--- a/pulsar/manager_endpoint_util.py
+++ b/pulsar/manager_endpoint_util.py
@@ -95,6 +95,9 @@ def submit_job(manager, job_config):
             jobs_directory = os.path.abspath(os.path.join(job_directory, os.pardir))
             command_line = command_line.replace('__PULSAR_JOBS_DIRECTORY__', jobs_directory)
             command_line = command_line.replace(os.path.join(jobs_directory, _get_galaxy_job_id(job_id)), job_directory)
+            for action in remote_staging['setup']:
+                if 'action' in action.keys() and 'contents' in action['action']:
+                    action['action']['contents'] = action['action']['contents'].replace(os.path.join(jobs_directory, _get_galaxy_job_id(job_id)), job_directory)
 
         # TODO: Handle __PULSAR_JOB_DIRECTORY__ config files, metadata files, etc...
         manager.touch_outputs(job_id, touch_outputs)
--- a/pulsar/manager_endpoint_util.py
+++ b/pulsar/manager_endpoint_util.py
@@ -13,12 +13,17 @@ from pulsar.managers import PULSAR_UNKNOWN_RETURN_CODE
 log = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


pulsar/manager_endpoint_util.py
```diff
@@ -85,9 +90,14 @@ def submit_job(manager, job_config):
...
 
         if job_config is not None:
+            job_id = job_config["job_id"]
             job_directory = job_config["job_directory"]
             jobs_directory = os.path.abspath(os.path.join(job_directory, os.pardir))
             command_line = command_line.replace('__PULSAR_JOBS_DIRECTORY__', jobs_directory)
+            command_line = command_line.replace(os.path.join(jobs_directory, _get_galaxy_job_id(job_id)), job_directory)
+            for action in remote_staging['setup']:
+                if 'action' in action.keys() and 'contents' in action['action']:
+                    action['action']['contents'] = action['action']['contents'].replace(os.path.join(jobs_directory, _get_galaxy_job_id(job_id)), job_directory)
```
jmchilton: Thank you so much for the contribution! I like parts of it but I am working so hard on making Pulsar's handling of inputs, paths, and command-line generation more structured - just rewriting everything after the fact on the Pulsar side like this seems really error prone and heading in the opposite direction. Let me think about this and the underlying issue - but if you can avoid using ``__PULSAR_JOBS_DIRECTORY__`` I would recommend it.

pulsar/managers/base/__init__.py
```diff
@@ -38,7 +38,7 @@
 ID_ASSIGNER = {
     # Generate a random id, needed if multiple
     # Galaxy instances submitting to same Pulsar.
-    'uuid': lambda galaxy_job_id: uuid4().hex,
+    'uuid': lambda galaxy_job_id: "{}-{}".format(galaxy_job_id, uuid4().hex),
```
jmchilton: Can you create a new type - ``job_plus_uuid`` for this? I feel like ``uuid`` should stick with uuids as job IDs and maybe just document what other parameters it doesn't work with for now?

bgruening:
@jmchilton any chance to get the UUID stuff into the next release? Not urgent, just asking.

jmchilton:
Thanks again for the fix but I think I prefer https://github.com/galaxyproject/galaxy/pull/8718 because I believe it will be more correct because it generates the correct paths at the outset and generates command lines, configfiles, etc... with the correct path instead of doing re-writing after the fact.

bgruening:
Thanks John for looking at it. @gmauro I guess we need to backport the Galaxy fix then to 19.09 for our instance.

PR #206, opened 2019-08-31
Add autodeploy to pulsar

bgruening:
@jmchilton I found the travis auto-deploy feature really useful.

https://docs.travis-ci.com/user/deployment/pypi/

What you need to do is the following:

```
gem install travis
/home/bag/.gem/ruby/2.5.0/bin/travis encrypt <your_passwd_here> --add deploy.password -r galaxyproject/pulsar
git commit .travis.yml -m 'add key'
```

If you have some special symbols in your password, then this might be helpful: https://docs.travis-ci.com/user/encryption-keys/#note-on-escaping-certain-symbols

The current configuration pushes to pypi as soon as you tag a release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,17 @@
 # TODO: test condor
 language: python
 dist: trusty
+
+deploy:
+  provider: pypi
+  user: jmchilton
+  skip_existing: true
+  password:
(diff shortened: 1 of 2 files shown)


jmchilton:
Pulsar needs to publish two wheels at the same time. The Makefile takes care of this and this wouldn't right? I like the idea maybe but start with Planemo maybe and update docs?

bgruening:
Imho this will build both packages. As is also "just" doing a `sdist bdist_wheels`, which will build both. But yes this PR is not at all important ... and the Makefile works. We could use the Makefile to archive the same I guess if you prefere ...

jmchilton:
It isn't the two artifacts - it builds two separate projects. https://pypi.org/project/pulsar-app/ and https://pypi.org/project/pulsar-galaxy-lib/.

bgruening:
Yes, `pulsar_app-0.14.0-py2.py3-none-any.whl         pulsar-app-0.14.0.tar.gz                       pulsar_galaxy_lib-0.14.0-py2.py3-none-any.whl  pulsar-galaxy-lib-0.14.0.tar.gz`.

However, I can not test it easily and it is not needed at all.

PR #214, opened 2020-02-13
Unpin six

nsoranzo:
and require a version supported by latest virtualenv.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 webob
 psutil
 PasteDeploy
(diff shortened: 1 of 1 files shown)


nsoranzo:
The new `six` minimum version required by virtualenv is causing also the Travis job failures here, which will hopefully be fixed by https://github.com/tox-dev/tox/pull/1519 and a new tox release.

nsoranzo:
Fixed tox is out, Travis is green.

PR #215, opened 2020-02-21
Unpin six

nsoranzo:


nsoranzo:
@jmchilton Could you please make a PyPI release which includes this fix?

PR #229, opened 2020-08-10
Fix a unicode issue and polish a bit of variables

gmauro:
This PR fixes the "Exception: cannot use a string pattern on a bytes-like object" into the condor manager with py3.


bgruening:
Thanks @gmauro!

gmauro:
Super!

PR #231, opened 2020-08-11
Drop Python 2.7 support in standard transport, drop Python 2.7 tests and fix Python 3.7 wheel install test, general test debugging enhancements

natefoo:
It's not going to work on 2.7 much longer anyway as galaxy-lib no longer supports 2.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,10 @@ language: python
 dist: bionic
 matrix:
   include:
-  - python: 2.7
-    env: TOX_ENV=py27-lint
-  - python: 2.7
-    env: TOX_ENV=py27-lint-dist
-  - python: 2.7
+  - python: 3.7
+    env: TOX_ENV=py37-lint-dist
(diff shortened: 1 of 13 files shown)


natefoo:
We can also just leave Paste and PasteScript in the `requirements.txt`, but they're unnecessary in a webless install.

bgruening:
It seems we still have a paste-test that fails: https://travis-ci.org/github/galaxyproject/pulsar/jobs/717076970#L947

bgruening (commented):
> [testenv:py27-install-wheel]

Should that be removed as well?

natefoo:
> > [testenv:py27-install-wheel]
> 
> Should that be removed as well?

Yep, thanks.

bgruening:
Cool, thanks Nate!

jmchilton:
Looks good to me - thanks for all the work @natefoo !

PR #232, opened 2020-08-15
fix the PyYAML "load() deprecation" warning

gmauro:
Explicitly specify the FullLoader to avoid the warning during the boot-up. 
This is currently (PyYAML 5.1) the default loader called by yaml.load(input) (after issuing the warning).
See [PyYAML-yaml.load(input)-Deprecation](https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation)

### DIFF
--- a/pulsar/client/config_util.py
+++ b/pulsar/client/config_util.py
@@ -56,7 +56,7 @@ def __read_yaml(path):
         raise ImportError("Attempting to read YAML configuration file - but PyYAML dependency unavailable.")
 
     with open(path, "rb") as f:
-        return yaml.load(f)
+        return yaml.load(f, Loader=yaml.FullLoader)
 
 
 def __read_ini(path):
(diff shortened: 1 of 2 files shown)


bgruening:
Should we maybe just use `safe_load`?

gmauro:
I think not. The function is loading our own yaml file, that we trust.

pulsar/client/config_util.py
```diff
@@ -56,7 +56,7 @@ def __read_yaml(path):
         raise ImportError("Attempting to read YAML configuration file - but PyYAML dependency unavailable.")
 
     with open(path, "rb") as f:
-        return yaml.load(f)
+        return yaml.load(f, Loader=yaml.FullLoader)
```
bgruening, line 59: This should probably safe_load() ... @natefoo ?

natefoo:
@gmauro We use `yaml.safe_load()` all over Galaxy - you're right that this is our own trusted yaml file, but we have no need to load anything other than basic data structures from it, so safe_yaml is still preferable.

gmauro:
@natefoo @bgruening I agree with your observation.

bgruening (approved):

natefoo (approved):

natefoo:
Thanks @gmauro!

Issue #233, opened 2020-09-16
Make a final Python 2.7 release of 0.13?


natefoo:
Changes in the master branch now (pre-0.14) fully break support for Python 2.7, but even installing 0.13 now will fail since it pulls the 20.5 versions of Galaxy libraries, which contain 2.7-breaking syntax.

Perhaps we should make one final 0.13 point release with the galaxy deps pinned to <20.5 so that it's still installable (for now) with 2.7? I know I can at least run the py27-install-wheel test with those changes, although other unpinned deps could be problematic as well (now, or some time in the future).

gmauro:
Sounds like a good idea to me.

natefoo:
Released as 0.13.1.

PR #236, opened 2020-09-16
Pin direct dependency versions and do a final Python 2.7 release.

natefoo:
This won't merge and we'll want to revert the pins and so forth, just creating this for the record.

As discussed in #233.


natefoo:
This is released (for the record, I screwed up the release so this version does not appear on Test PyPI (and the version of 0.14.0 on Test PyPI will be forever incorrect, so that's cool)).

PR #238, opened 2020-11-12
Better error handling when setting up environment statements

natefoo:
Also always convert env var value to a string.


natefoo:
Also always convert env var value to a string.

jmchilton:
Closing in favor the now merged #504. Thanks for the fix!

PR #246, opened 2021-02-15
Set resource limits and walltime for k8s coexecution pods

natefoo:
Marius van den Beek @mvdbeek 11:06
> Before replicating all this between Galaxy and pulsar in slightly different ways it might be a good idea to converge this in pulsar, and maybe make all of this templateable ?
Longer term ofc

Nuwan Goonasekera @nuwang 11:11
> Yes, that would be good. There was a preliminary attempt at this which was abandoned pending the shift to Pulsar, but perhaps the template etc. can still be used? galaxyproject/galaxy#10714

Nate Coraor @natefoo 11:21
> pfft copy paste is the best method ;D

Nate Coraor @natefoo 11:29
> (yes, you're right, but if this works I'm PRing it anyway as I need it to work in production and we can refactor it later)

### DIFF
--- a/pulsar/client/client.py
+++ b/pulsar/client/client.py
@@ -468,6 +468,7 @@ class MessageCoexecutionPodJobClient(BaseMessageJobClient):
         params = self.destination_params
 
         pulsar_container_image = self.pulsar_container_image
+        pulsar_container_resources = self._pulsar_container_resources(params)
 
         job_directory = self.job_directory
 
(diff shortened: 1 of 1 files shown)


jmchilton:
Oh dear - this is going to spiral out of control. Q2 we really need to have synchronizing Pulsar and Kubernetes runner code as an objective.

natefoo:
Q2 all Galaxy jobs execute through Pulsar, problem solved!

PR #247, opened 2021-03-02
Small fix for starting pulsar in webless mode

gregvonkuster:
When pulsar is configured with --mq, running this command:
```
pulsar --mode webless
```
will throw this exception:
```
FileNotFoundError: [Errno 2] No such file or directory: '~/pulsar/server.ini.sample'
```
This occurs because ```ini_path``` is initialized to ```~/pulsar/server.ini.sample``` here https://github.com/galaxyproject/pulsar/blob/master/pulsar/main.py#L254, and the sample config file is not created by pulsar-config when passing the --mq option.


jmchilton:
Looks like this is now superseded by other fixes and conflicts - sorry for not getting to it sooner 😿. Thanks for the contribution.

PR #257, opened 2021-07-07
Support globs in from_work_dir outputs

### LINKED ISSUE #239: Glob/wildcard in from_work_dir outputs doesn't work
e.g. [in cutadapt](https://github.com/galaxyproject/tools-iuc/blob/41371d219eaafc811eadf7646a7f0ecf002e4239/tools/cutadapt/cutadapt.xml#L242) it attempts to collect `out1*` and `out1.fa` exists:

```
$ ls -l /jetstream/scratch0/main/jobs/1/working/out1*
-rw-rw-rw-. 1 galaxy galaxy 3390441465 Jan 11 22:46 /jetstream/scratch0/main/jobs/1/working/out1.fa
```

But collection fails:

```
Jan 11 23:01:12 jetstream-iu0.galaxyproject.org pulsar[1521]: 2021-01-11 23:01:12,696 WARNI [pulsar.managers.staging.post][[manager=jetstream_iu]-[action=postprocess]-[job=1]] Failures collecting results [Exception('Attempt to post file /jetstream/scratch0/main/jobs/1/working/out1* to URL https://galaxy/_job_files?job_id=c0ffee&job_key=deadbeef&path=/galaxy-repl/main/files/000/dataset_2.dat&file_type=output_workdir, but file does not exist.',)]
```

natefoo:
Fixes #239.

I am not sure about what the duplication is for in `client/job_directory.py` and what to do for that case. Also, no idea how we'll get the tool profile handling down to `calculate_path()`, but that is a problem for future Nate?

### DIFF
--- a/pulsar/managers/base/__init__.py
+++ b/pulsar/managers/base/__init__.py
@@ -2,7 +2,6 @@
 Base Classes and Infrastructure Supporting Concret Manager Implementations.
 
 """
-from collections import deque
 import errno
 import logging
 import os
@@ -16,7 +15,6 @@ from os import getenv
 from os import walk
 import json
 from uuid import uuid4
-import posixpath
 from shutil import rmtree
 
 import six
@@ -26,7 +24,6 @@ from pulsar.managers import ManagerInterface
 from pulsar.client.job_directory import (
     RemoteJobDirectory,
     get_mapped_file,
-    verify_is_in_directory,
 )
 
 JOB_DIRECTORY_INPUTS = "inputs"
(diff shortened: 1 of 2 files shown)


pulsar/managers/base/__init__.py
```diff
@@ -384,6 +385,15 @@ def get_mapped_file(directory, remote_path, allow_nested_files=False, local_path
         if mkdir and not local_path_module.exists(local_directory):
```
jmchilton, line 385: Is there any difference in these methods? If not - don't you want to delete this one and import that other one?

natefoo: I assumed you had some good reason for it being copy/pasted. 😂

jmchilton (approved):
Bleh...

PR #258, opened 2021-07-08
Allow from_work_dir output stage out failure

### LINKED ISSUE #211: Missing stderr/stdout on pulsar jobs
Not sure if this is a duplicate of https://github.com/galaxyproject/pulsar/issues/158 but I recently ran some circos jobs on EU (which go through pulsar). All of the jobs errored out. https://usegalaxy.eu/u/helena-rasche/h/circos-multi-plot

- "Remote job server indicated a problem running or monitoring this job."
- No stderr
- No stdout

I had to contact the admins to help debug this issue and they could look at the job logs directly. It would be super useful for me if the job stderr/out were present, since any errors during training which uses EU will only have this cryptic message.

Thanks for the tool debugging help @gmauro @bgruening

natefoo:
When a tool fails it's often not going to create `from_work_dir` outputs. Currently this causes Pulsar to send a `failed` state to Galaxy due to stageout failures, and this makes diagnosing job failure causes much less apparent to the user.

i.e. this change turns this unhelpful message:

Screen Shot 2021-07-08 at 6 16 45 PM

into this:

Screen Shot 2021-07-08 at 6 16 52 PM

(where "Tool failed for reasons" is the actual error message generated by my very fine failure testing tool).

Fixes #211.

### DIFF
--- a/pulsar/client/staging/down.py
+++ b/pulsar/client/staging/down.py
@@ -167,7 +167,7 @@ class DownloadExceptionTracker(object):
         except Exception as e:
             if allow_failure:
                 log.warning(
-                     "Allowed failure in postprocessing, will not force job failure but generally indicates a tool"
+                    "Allowed failure in postprocessing, will not force job failure but generally indicates a tool"
                     f" failure: {e}")
             else:
                 self.collection_failure_exceptions.append(e)
--- a/pulsar/client/staging/down.py
+++ b/pulsar/client/staging/down.py
@@ -143,7 +143,7 @@ class ResultsCollector(object):
         # name is the 'name' of the file on the Pulsar server (possible a relative)
         # path.
         collected = False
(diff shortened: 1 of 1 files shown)


hexylena:
OMG this is going to save so much time.

pulsar/client/staging/down.py
```diff
@@ -143,7 +143,7 @@ def _attempt_collect_output(self, output_type, path, name=None):
         # name is the 'name' of the file on the Pulsar server (possible a relative)
         # path.
         collected = False
-        with self.exception_tracker():
+        with self.exception_tracker(allow_failure=_allow_collect_failure(output_type)):
             action = self.action_mapper.action({"path": path}, output_type)
             if self._collect_output(output_type, action, name):
```
jmchilton, line 148: Can you move this behavior into collect_output for these outputs? This seems like an odd level to handle it at right? Like - if this isn't an exception condition don't throw it as one?

PR #260, opened 2021-07-09
Transfer entire tool directory instead of picking apart the generated command line to find tool files to transfer

### LINKED ISSUE #113: Staging directory can miss helper scripts etc
e.g. https://github.com/peterjc/galaxy_mira/issues/1 where the tool folder includes multiple files, including:
- `mira4_de_novo.xml` <- one of the tool definitions
- `mira4.py` <-- wrapper script called in XML file, appears to be staged
- `mira4_make_bam.py` <-- imported from wrapper script, not staged

On that issue @jmchilton wrote:

_Pulsar doesn't know how to see the required files that need to be transferred - tool authors need to be able to mark these up in the tool XML but it is currently not implemented (galaxyproject/galaxy#1609). This is urgent and is something I will do this summer I think - but probably not until after the ISMB._

I think in the short term it would be safer to default to staging _all_ the files next to the tool XML, as historically Galaxy tool authors have been able to assume they will be present in the same folder.

natefoo:
Many tools work by including/importing things from the tool directory in scripts that are not themselves called in the command. This transfers the entire tool directory (with the exception of `tool-data`, `test-data`, and `.hg`) instead. Maybe more than we need but hopefully shouldn't be *too* painful.

Fixes #113.

### DIFF
--- a/pulsar/client/staging/up.py
+++ b/pulsar/client/staging/up.py
@@ -471,7 +471,7 @@ class TransferTracker(object):
             remote_name = self.path_helper.remote_name(relpath(directory_file_path, directory))
             self.handle_transfer_path(directory_file_path, type, name=remote_name)
 
-    def handle_transfer_source(self, source, type, name=None, contents=None):
+    def handle_transfer_source(self, source, type, name=None, contents=None):  # noqa: C901
         action = self.__action_for_transfer(source, type, contents)
 
         if action.staging_needed:
--- a/pulsar/client/staging/up.py
+++ b/pulsar/client/staging/up.py
@@ -10,6 +10,8 @@ from os.path import (
 )
 from re import escape, findall
 
+from galaxy.util.path import safe_walk
+
 from ..action_mapper import FileActionMapper
 from ..action_mapper import MessageAction
 from ..action_mapper import path_type
@@ -98,7 +100,7 @@ class FileStager(object):
 
         # Setup job inputs, these will need to be rewritten before
         # shipping off to remote Pulsar server.
-        self.job_inputs = JobInputs(self.command_line, self.config_files)
+        self.job_inputs = JobInputs(self.command_line, self.config_files, self.tool_dir)
 
         self.action_mapper = FileActionMapper(client)
 
@@ -114,7 +116,7 @@ class FileStager(object):
             self.job_directory,
         )
 
-        self.__initialize_referenced_tool_files()
+        self.__initialize_tool_directory()
         if self.rewrite_paths:
             self.__initialize_referenced_arbitrary_files()
 
@@ -178,18 +180,14 @@ class FileStager(object):
         else:
             return None
 
-    def __initialize_referenced_tool_files(self):
-        # Was this following line only for interpreter, should we disable it of 16.04+ tools
-        self.referenced_tool_files = self.job_inputs.find_referenced_subfiles(self.tool_dir)
-        # If the tool was created with a correct $__tool_directory__ find those files and transfer
+    def __initialize_tool_directory(self):
+        # Find and transfer all files and subdirs of $__tool_directory__
+        # TODO: What is this for? Was used in the previous "referenced" version of this method
         new_tool_directory = self.new_tool_directory
         if not new_tool_directory:
             return
 
-        for potential_tool_file in self.job_inputs.find_referenced_subfiles(new_tool_directory):
-            local_file = potential_tool_file.replace(new_tool_directory, self.tool_dir)
-            if exists(local_file):
-                self.referenced_tool_files.append(local_file)
+        self.tool_files = self.job_inputs.find_tool_files()
 
     def __initialize_referenced_arbitrary_files(self):
         referenced_arbitrary_path_mappers = dict()
@@ -207,8 +205,8 @@ class FileStager(object):
             self.arbitrary_files.update(unstructured_map)
 
     def __upload_tool_files(self):
-        for referenced_tool_file in self.referenced_tool_files:
-            self.transfer_tracker.handle_transfer_path(referenced_tool_file, path_type.TOOL)
+        for tool_file in self.tool_files:
+            self.transfer_tracker.handle_transfer_path(tool_file, path_type.TOOL)
 
     def __upload_job_directory_files(self):
         for job_directory_file in self.job_directory_files:
@@ -333,7 +331,7 @@ class JobInputs(object):
     >>> tf = tempfile.NamedTemporaryFile()
     >>> def setup_inputs(tf):
     ...     open(tf.name, "w").write(u'''world /path/to/input '/path/to/moo' "/path/to/cow" the rest''')
-    ...     inputs = JobInputs(u"hello /path/to/input", [tf.name])
+    ...     inputs = JobInputs(u"hello /path/to/input", [tf.name], None)
     ...     return inputs
     >>> inputs = setup_inputs(tf)
     >>> inputs.rewrite_paths(u"/path/to/input", u'C:\\input')
@@ -356,10 +354,16 @@ class JobInputs(object):
     False
     >>> tf.close()
     """
+    IGNORE_TOOL_FILES = set([
+        'tool-data',
+        'test-data',
+        '.hg',
+    ])
 
-    def __init__(self, command_line, config_files):
+    def __init__(self, command_line, config_files, tool_directory):
         self.command_line = command_line
         self.config_files = {}
+        self.tool_directory = tool_directory
         for config_file in config_files or []:
             config_contents = _read(config_file)
             self.config_files[config_file] = config_contents
@@ -387,6 +391,15 @@ class JobInputs(object):
 
         pattern = r'''[\'\"]?(%s%s[^\s\'\"]+)[\'\"]?''' % (escape(directory), escape(sep))
         return self.find_pattern_references(pattern)
+
+    def __ignore_tool_files(self, filename):
+        return filename in self.IGNORE_TOOL_FILES
+
+    def find_tool_files(self):
+        tool_files = []
+        for (dirpath, dirnames, filenames) in safe_walk(self.tool_directory):
+            tool_files.extend(join(dirpath, filename) for filename in filenames if not self.__ignore_tool_files(filename))
+        return tool_files
 
     def path_referenced(self, path):
         pattern = r"%s" % path
@@ -482,7 +495,10 @@ class TransferTracker(object):
                 if not name:
                     # TODO: consider fetching this from source so an actual input path
                     # isn't needed. At least it isn't used though.
-                    name = basename(path)
+                    if type == 'tool':
+                        name = path.replace(self.job_inputs.tool_directory + sep, '')
+                    else:
+                        name = basename(path)
                 self.__add_remote_staging_input(action, name, type)
 
                 def get_path():
--- a/pulsar/client/staging/up.py
+++ b/pulsar/client/staging/up.py
@@ -181,8 +181,9 @@ class FileStager(object):
             return None
 
     def __initialize_tool_directory(self):
+        # This following line is only for interpreter, we should disable it for 16.04+ tools
+        self.tool_files = self.job_inputs.find_referenced_subfiles(self.tool_dir)
         # Find and transfer all files and subdirs of $__tool_directory__
-        # TODO: What is this for? Was used in the previous "referenced" version of this method
         new_tool_directory = self.new_tool_directory
         if not new_tool_directory:
             return
--- a/pulsar/client/staging/up.py
+++ b/pulsar/client/staging/up.py
@@ -207,7 +207,8 @@ class FileStager(object):
 
     def __upload_tool_files(self):
         for tool_file in self.tool_files:
-            self.transfer_tracker.handle_transfer_path(tool_file, path_type.TOOL)
+            name = self.job_inputs.relative_file_name(self.tool_dir, tool_file)
+            self.transfer_tracker.handle_transfer_path(tool_file, path_type.TOOL, name=name)
 
     def __upload_job_directory_files(self):
         for job_directory_file in self.job_directory_files:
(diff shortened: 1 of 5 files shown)


natefoo:
Quick someone click merge before @jmchilton sees this

jmchilton:
No.

pulsar/client/staging/up.py
```diff
@@ -482,7 +496,10 @@ def get_path():
                 if not name:
                     # TODO: consider fetching this from source so an actual input path
                     # isn't needed. At least it isn't used though.
-                    name = basename(path)
+                    if type == 'tool':
```
jmchilton, line 499: No.

pulsar/client/staging/up.py
```diff
@@ -388,6 +393,15 @@ def find_referenced_subfiles(self, directory):
         pattern = r'''[\'\"]?(%s%s[^\s\'\"]+)[\'\"]?''' % (escape(directory), escape(sep))
         return self.find_pattern_references(pattern)
 
+    def __ignore_tool_files(self, filename):
+        return filename in self.IGNORE_TOOL_FILES
+
+    def find_tool_files(self):
```
jmchilton, line 399: If you can figure out how ``handle_transfer_source`` should actually work without that crazy hack then we could maybe do this by having a list of tool IDs that we do this for? The way we do with Galaxy lib exposure. No reason to break all of this and do all these unneeded transfers just for a handful of tools.

natefoo:
> No.

Boooo.

pulsar/client/staging/up.py
natefoo: Which part of it does "crazy hack" refer to? Sending everything for every tool?

i.e. is it ok to send everything for tools that we can determine need this treatment?

jmchilton: > i.e. is it ok to send everything for tools that we can determine need this treatment?

Yes - this is fine - I always assumed we'd do this with file actions but if you want to do it this way I won't stand in your way.

The hack I'm referring to is the line in I said No to "pulsar/client/staging/up.py" - that just cannot be the right way to handle it. The inputs to that function aren't being setup properly or we're not handling nested files properly or something if that is the fix needed.

natefoo:
What if we add a tool tag where wrapper authors can specify "extra tool files" in a new tool schema version. Tools newer than that version can then use the original "referenced tool files" method once support for the extended list is implemented in Pulsar.

For old tool versions, admins can either enable this PR's "send everything" functionality on a per-tool basis, or a "do it for all tools" switch? AU and EU really need this to work (and honestly it would help me too) but finding all the cases may be kinda painful (or maybe it won't? AU already has a pretty extensive list compiled).

pulsar/client/staging/up.py
natefoo: Nevermind, I gather than "crazy hack" refers to your comment in `handle_transfer_source`.

jmchilton:
If you provide a simple list - admins will rapidly add to it. I'd prefer not to switch the behavior to everything by default.  AU and EU are the main consumers of Pulsar - if they're using a non-standard option - the standard options won't be well tested.

pulsar/client/staging/up.py
natefoo: No to hacking the path (line below) or checking `type == 'tool'`? Because you conditional `type == 'tool'` down below in the same method.

pulsar/client/staging/up.py
jmchilton: Yes - this is fine behavior if it is intentional.

natefoo: Whoops, deleted my comments after I figured out what you meant. 😂 

Thanks for clarifying.

Slugger70: Could we perhaps detect extraneous files in the tool directory that aren't there in 95% of cases. i.e. If there is a helper script then we copy everything across. Otherwise nothing?

Slugger70: What I mean is - If the current command line detection returns nothing then we copy nothing to pulsar, but if it returns anything or finds a `__tool_directory__` in the command line then it copies everything . I hope this makes sense.

hexylena:
If we were to consider the everything approach:
- 99.4% of all tools on EU have tool directories smaller than 5.7MB 
- 94.5% of the above 99% are <1.4 MB
- Together it totals around 1.3GB of tools/etc (excluding `.hg`, `tool-data`, `test-data`, and `static/images`)
- 

If we were to cache any tools that were run using pulsar's caching feature, I reckon this would be miniscule (even uncached) relative to the size of data we push through pulsar. 

Here's a subsample of a table querying for jobs with a runner like pulsar, cross-referenced through jtid/jtod (not DCs) for their transfer volume:

```
   month    | AU sent  | AU recv | EU sent | EU recv
------------+----------+---------+---------+---------
 2021-01-01 | 6727 GB  | 344 GB  | 31 GB   | 3906 MB
 2021-02-01 | 16 TB    | 926 GB  | 80 GB   | 11 GB
 2021-03-01 | 17 TB    | 1089 GB | 52 GB   | 7169 MB
 2021-04-01 | 27 TB    | 4447 GB | 119 GB  | 29 GB
 2021-05-01 | 20 TB    | 4641 GB | 140 GB  | 83 GB
 2021-06-01 | 13 TB    | 2942 GB | 1979 MB | 2228 MB
 2021-07-01 | 4503 GB  | 689 GB  | 130 GB  | 1677 MB
```

And here's the queries I ran for the tool dirs.

```
$ rsync -avr repos --exclude=tool-data --exclude=test-data --exclude=.hg --exclude=static/images /tmp/testing
$ cd /tmp/testing/repos/
$ du -m -d 3 | egrep '/[a-f0-9]{5,}'  | cut -f 1 | histogram.py --percentage
# NumSamples = 6735; Min = 1.00; Max = 48.00
# Mean = 1.107944; Variance = 1.011214; SD = 1.005591; Median 1.000000
# each ∎ represents a count of 89
    1.0000 -     5.7000 [  6697]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎ (99.44%)
    5.7000 -    10.4000 [    29]:  (0.43%)
   10.4000 -    15.1000 [     4]:  (0.06%)
   15.1000 -    19.8000 [     1]:  (0.01%)
   19.8000 -    24.5000 [     1]:  (0.01%)
   24.5000 -    29.2000 [     0]:  (0.00%)
   29.2000 -    33.9000 [     2]:  (0.03%)
   33.9000 -    38.6000 [     0]:  (0.00%)
   38.6000 -    43.3000 [     0]:  (0.00%)
   43.3000 -    48.0000 [     1]:  (0.01%)

$ du -m -d 3 | egrep '/[a-f0-9]{5,}'  | cut -f 1 | histogram.py --percentage  --max=5
# NumSamples = 6735; Min = 1.00; Max = 5.00
# 38 values outside of min/max
# Mean = 1.107944; Variance = 1.011214; SD = 1.005591; Median 1.000000
# each ∎ represents a count of 84
    1.0000 -     1.4000 [  6363]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎ (94.48%)
    1.4000 -     1.8000 [     0]:  (0.00%)
    1.8000 -     2.2000 [   330]: ∎∎∎ (4.90%)
    2.2000 -     2.6000 [     0]:  (0.00%)
    2.6000 -     3.0000 [     1]:  (0.01%)
    3.0000 -     3.4000 [     0]:  (0.00%)
    3.4000 -     3.8000 [     0]:  (0.00%)
    3.8000 -     4.2000 [     3]:  (0.04%)
    4.2000 -     4.6000 [     0]:  (0.00%)
    4.6000 -     5.0000 [     0]:  (0.00%)
```

hexylena:
Looking at those recent numbers in context of how many jobs were run since that's important:

```
gxadmin query pulsar-gb-transferred --bymonth --human
   month    |   sent   |  recv   | job_count
------------+----------+---------+-----------
 2021-02-01 | 16 TB    | 926 GB  |     29164
 2021-03-01 | 17 TB    | 1089 GB |     56396
 2021-04-01 | 27 TB    | 4447 GB |     53568
 2021-05-01 | 20 TB    | 4641 GB |     44198
 2021-06-01 | 13 TB    | 2942 GB |     32492
 2021-07-01 | 4527 GB  | 689 GB  |      9278
```

that's coming out to about 500MB out per job which isn't very big (but I haven't included dataset collections because that's difficult). I don't think this completely kills the idea of "transfer everything" because the tool directories are incredibly cacheable.

Anyway this prompted me to re-check the repo sizes in kb rather than MB, and yeah. 

```
(sdf) root@sn04:/tmp/testing/repos$ du -k -d 3 | egrep '/[a-f0-9]{5,}'  | cut -f 1 | histogram.py --percentage -b 30
# NumSamples = 6735; Min = 4.00; Max = 48932.00
# Mean = 189.259094; Variance = 1195869.364644; SD = 1093.558121; Median 32.000000
# each ∎ represents a count of 88
    4.0000 -  1634.9333 [  6633]: ∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎∎ (98.49%)
```

pulsar/client/staging/up.py
natefoo: I *think* it should be pretty easy to limit this to tools that reference `$__tool_directory__` and tools that set `<command interpreter="">` by modifying `galaxy.tools:Tool` to store an attribute like `requires_tool_directory`, and then use that to determine which method should be used. Would that be reasonable to you, @jmchilton ?

natefoo: Scratch that, we're working on the list.

jmchilton: It is a lot to ask but if we could turn the list into something like a list of file actions or regexes - that just default to ``**`` for the tools in the list I think that would be ideal.

- There will be tools that have big directories that we want to exclude stuff from and we can just make the list a dictionary.
- There will be tools that we need to capture the parent directory contents of - again we can just do that there also I think.
- Then we could expose tool XML constructs for overriding this list - which has been on the TODO list forever. We can take care of the long-term goals while getting these legacy things done. I'll try to help, I know this is a lot to ask and you're just trying to get shit working - sorry for being this way 😅.

jmchilton: Can you let me know when you have a WIP Galaxy side of this?

pulsar/client/staging/up.py
```diff
@@ -207,7 +207,8 @@ def __initialize_referenced_arbitrary_files(self):
 
     def __upload_tool_files(self):
         for tool_file in self.tool_files:
-            self.transfer_tracker.handle_transfer_path(tool_file, path_type.TOOL)
+            name = self.job_inputs.relative_file_name(self.tool_dir, tool_file)
```
jmchilton, line 210: This looks a lot better to my eye - I still don't understand why this needed for tools but not for say UNSTRUCTURED files (``__upload_arbitrary_files``). That is really odd to me - but at least the hack is more limited now.

jmchilton:
The physical size on disk of the directories is not the only concern here. The Pulsar setup message will explode based on the number of files also. There is a lot of overhead for each file transferred.

hexylena:
I hid my comments as outdated now that we've settled on the list. Thanks for the details, I hadn't realised they were part of the setup messages.

jmchilton:
I saw you hid your comments but I thought they were really valuable and I just wanted to add some context. Sorry for being a pain.

hexylena:
Oh, no worries! /gen
I'd really forgotten about the setup messages and discounted the overhead, it's good to hear from someone much more familiar with the codebase.

Ah the data's nice to have, I can unhide them then.

natefoo:
Ok, the start of the Galaxy side of things is in this branch on my fork:

https://github.com/natefoo/galaxy/tree/unreferenced-tool-files

There is a stub in the Pulsar runner where the information about the tool requiring unreferenced tool files will need to be conveyed to Pulsar through some as-yet-unwritten Pulsar client changes on this PR.

jmchilton:
Can I have a seed list of a few tools we want to target for fixes out of the box. And bonus points if the list includes some tools that no longer need these fixes but did at prior versions we know of.

hexylena:
Tool list:

- deseq2 could have it's workaround reverted, https://github.com/galaxyproject/tools-iuc/blob/master/tools/deseq2/deseq2.xml#L36
- spyboat, https://github.com/galaxyproject/tools-iuc/blob/e22821f51ae326b3696a9456cbc44f9982512a52/tools/spyboat/spyboat.xml

jmchilton:
Transferring the whole tool directory is hackier than those tool hacks IMO. The tools are now referencing the required files and not just counting on them being there. But I do appreciate the links since I can tell older deseq2 is broken and so my new PR will fix those older versions I think.

Regardless of which hack is hackier, I think we will all be happier when tools declare their required files explicitly. I think the framework for that is outlined in https://github.com/galaxyproject/galaxy/pull/12250. Which I think is a better approach to the Galaxy side of this - ClientJobDescription will now have an optional ``tool_directory_required_files`` object parsed from the tool XML or from dicts for specific tool versions in tools/__init__.py for legacy tools. 

I outlined in the PR how to build a RequiredFiles (https://github.com/galaxyproject/galaxy/pull/12250/files#diff-52a44fbf0aeca839d987aba8f9bc29f266b51cffab9213473dd20e3979ce0cc6R460) that just transfers the whole directory but based on the hacks linked above I think in most cases it will just be a file or two that is needed. This sort of matches my intuition.

I think that RequiredFiles object should have everything we need to rewrite the file iteration walk in this PR and I think if this object is not set - we should just retain the original behavior before this PR.

pulsar/client/staging/up.py
natefoo: You mean this?

https://github.com/natefoo/galaxy/tree/unreferenced-tool-files

PR #262, opened 2021-07-14
Fix Galaxy composite input references.

jmchilton:
Pulsar side of a fix for https://github.com/galaxyproject/galaxy/issues/12256 - I think the only thing that would need to happen Galaxy side is a version upgrade of the pulsar-galaxy-lib post release of this fix.

### DIFF
--- a/docs/developing.rst
+++ b/docs/developing.rst
@@ -25,7 +25,7 @@ are fine.
     password:<password>
     
     [test]
-    repository:https://testpypi.python.org/pypi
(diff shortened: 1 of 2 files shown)


mvdbeek:
Do you want to push this to my PR branch with the tests ? That should tell you if it worked.

jmchilton:
Well lets just merge your branch right.

jmchilton:
I don't think I realize it tested the client also - that is very nice! Thanks so much.

PR #265, opened 2021-07-16
Is this needed for branching to work?

jmchilton:
I was trying to parameterize job conf based on branch and I didn't see how this would work without this.


mvdbeek (approved):
Oops!

Issue #263, opened 2021-07-15
Support Expression Tools in Pulsar


jmchilton:
Need to transfer the Python script - perhaps as a config file or tool file.

jmchilton:
exec_before_job writes files Pulsar doesn't know about and don't seem to respect compute environments:

```
        local_working_directory = param_dict["__local_working_directory__"]
        expression_inputs_path = os.path.join(local_working_directory, ExpressionTool.EXPRESSION_INPUTS_NAME)
```

The cd ../ is a bit questionable also. 

```
        self.command = f"cd ../; {expressions.EXPRESSION_SCRIPT_CALL}"
```

PR #266, opened 2021-07-16
Use tool classes to only test remote Galaxy tools.

jmchilton:


jmchilton:
That cut down the errors nicely. Arguably valid errors because of #263 for instance but I'd rather get the expected real tools passing first - maybe get PRs failing on errors here.

PR #267, opened 2021-07-16
Attempt to fix local manager not respecting command exit codes.

jmchilton:
The exit codes aren't being respected in the framework tests but they must work in production right? Someone would have created an issue - so I assume it is the "local" runner equivalent and this TODO seems like it would cause the same problem.


jmchilton:
```
galaxy.tool_util.verify.interactor.JobOutputsError: Expected job to fail but Galaxy indicated the job successfully completed.
```

Not fixed.

Issue #93, opened 2016-01-06
Handle galaxy.json


jmchilton:
- [ ] Write the file with new paths.
- [ ] Send outputs defined within back to Galaxy.

PR #269, opened 2021-07-16
Implement dynamic file sources abstraction for parsing files to transfer from galaxy.jsons

jmchilton:
A matching Galaxy PR needs to be done after the client is published that records the expected location of the tool provided metadata file and the type.

Matching client update in the integration tests.

https://github.com/galaxyproject/pulsar/pull/269/files#diff-8d13dd3d59bfdedd1272c7a1bd9f3b76f24224ca80a5c1185775c905e250827bR270

Fixes https://github.com/galaxyproject/pulsar/issues/93


natefoo:
Does this mean Pulsar now fully supports the extended metadata strategy?

PR #276, opened 2021-07-20
Another attempt at replacing Galaxy's pulsar wheel for CI

jmchilton:
I think the other version didn't work properly #275.


mvdbeek (approved):
Thanks, I didn't even notice that the previous variant didn't work.

jmchilton:
Well I had the advantage of being frustrated that #267 didn't work to fix an issue I was having. The audacity of the tests to imply anything I did was less than perfect was too much to tolerate.

With this fix though we're down to just 11 failing tests in dev down from 50 failures in the original PR that you added these tests in!

PR #279, opened 2021-07-21
Rev job script code against latest Galaxy dev.

jmchilton:


pulsar/managers/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
```diff
@@ -4,7 +4,8 @@ $headers
 
 _galaxy_setup_environment() {
     local _use_framework_galaxy="$1"
-    _GALAXY_JOB_HOME_DIR="$working_directory/../home"
+    _GALAXY_JOB_DIR="$working_directory"
```
jmchilton, line 7: I guess this is not right at all in a Pulsar context where I think working_directory is the tool working directory not the job directory.

pulsar/managers/util/job_script/DEFAULT_JOB_FILE_TEMPLATE.sh
```diff
@@ -4,7 +4,8 @@ $headers
...
+    _GALAXY_JOB_HOME_DIR="$working_directory/home"
```
mvdbeek, line 8: But then home will then be right in the working directory. We don't do that in Galaxy and might impact recursively collecting outputs, for instance.

PR #282, opened 2021-08-10
avoid using getcwd() in cli module pattern matching

martenson:
I think this did not work in most of the cases due to the getcwd() call.

### DIFF
--- a/pulsar/managers/util/cli/__init__.py
+++ b/pulsar/managers/util/cli/__init__.py
@@ -24,7 +24,7 @@ class CliInterface(object):
         """
         """
         def __load(module_path, d):
-            module_pattern = join(join(getcwd(), code_dir, *module_path.split('.')), '*.py')
+            module_pattern = join(join(dirname(__file__), module_path.split('.')[-1]), '*.py')
             for file in glob(module_pattern):
                 if basename(file).startswith('_'):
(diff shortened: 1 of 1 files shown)


pulsar/managers/util/cli/__init__.py
```diff
@@ -24,7 +24,7 @@ def __init__(self, code_dir='lib'):
         """
         """
         def __load(module_path, d):
-            module_pattern = join(join(getcwd(), code_dir, *module_path.split('.')), '*.py')
+            module_pattern = join(join(dirname(__file__), module_path.split('.')[-1]), '*.py')
```
mvdbeek, line 27: I think the implementation of the pulsar variant and the Galaxy variant have diverged.
I think you could copy https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/util/cli/__init__.py over and that should fix it too ? In either case we should unify these two before we diverge further.

Might even be worth to delete the Galaxy variant and only keep the pulsar code.

martenson: related https://github.com/galaxyproject/galaxy/pull/10189

jmchilton:
Superseded by d254d8c - thanks!

PR #288, opened 2021-10-21
Remove Python <=3.5 code

nsoranzo:
Using `ack --type=python -f | xargs pyupgrade --py3-plus`

Also:
- Remove dependency on `six`
- Fix some issues reported by flake8-bugbear
- Import `getfullargspec` directly from `inspect`


mvdbeek (approved):
Thank you @nsoranzo!

PR #307, opened 2022-10-03
Fix https://github.com/galaxyproject/pulsar/pull/299.

jmchilton:
Last Galaxy error I think.

Fixes error with https://github.com/galaxyproject/galaxy/pull/14748


mvdbeek (approved):
Ahhh, sorry about that!

PR #278, opened 2021-07-21
Fix memory_statement.log being written to working directory.

jmchilton:
Fixes #277

PR #309, opened 2022-10-12
Re-apply fix from #278

jmchilton:
Got lost in sync-ing with Galaxy maybe?


bgruening:
@mvdbeek this looks also relevant to your PR.

mvdbeek:
yes, that is part of #318 too

mvdbeek:
I didn't exactly copy over the updated script from Galaxy, since the current version uses dataclasses and so isn't compatible with python 3.6 ... in either case we'd have to update the PR.

PR #314, opened 2023-02-03
If `manager` is specified in the app conf and --manager is supplied, make the single manager have the supplied name

### LINKED ISSUE #313: Coexecution jobs don't work with non-default managers
A non-default manager is necessary in this case because it's how we route messages to the correct Pulsar via AMQP.

Upon upgrading to usegalaxy.org, 23.0, Pulsar Kubernetes runner jobs are failing with the following error (where `tacc_k8s` is the manager defined in the [runner plugin](https://github.com/galaxyproject/usegalaxy-playbook/blob/421507decc91a18f75b8727db19e6a85dc4a8b42/env/common/templates/galaxy/config/job_conf.yml.j2#L164) and is [the destination id](https://github.com/galaxyproject/usegalaxy-playbook/blob/421507decc91a18f75b8727db19e6a85dc4a8b42/env/common/templates/galaxy/config/job_conf.yml.j2#L291):

```pytb
Traceback (most recent call last):
  File "/usr/local/bin/pulsar-submit", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.7/site-packages/pulsar/scripts/submit.py", line 20, in main
    run_server_for_job(args)
  File "/usr/local/lib/python3.7/site-packages/pulsar/scripts/submit_util.py", line 29, in run_server_for_job
    manager, app = manager_from_args(config_builder)
  File "/usr/local/lib/python3.7/site-packages/pulsar/scripts/submit_util.py", line 94, in manager_from_args
    manager = pulsar_app.managers[manager_name]
KeyError: 'tacc_k8s'
```

usegalaxy.org had been running a [custom coexecution image](https://quay.io/repository/natefoo/pulsar-pod-staging) and unfortunately I no longer have any recollection as to why or how that image was built. But the version of Pulsar appears to be from somewhere around d9b710231cba65555285d38ae4da139ef5421395 and I can't see any differences that suggest I manually hacked a fix. But in local testing I can reproduce this error all the way back to very old 0.14.x versions, so I have been unable to figure out how this was working until now. This image might still work if not for the fact that the client is adding `--wait` to the `pulsar-submit` args and the version of `pulsar-submit` in that image doesn't have `--wait`.

It may be possible/correct to set a pulsar app config on the Galaxy side that explicitly defines the `tacc_k8s` manager. The app conf generated for `pulsar-submit`'s `--app_conf_base64` option for these jobs contains:

```json
{
  "staging_directory": "/pulsar_staging",
  "message_queue_url": "amqp://user:pass@host:5671//main_pulsar?ssl=1",
  "manager": {
    "type": "coexecution",
    "monitor": "background"
  },
  "persistence_directory": "/pulsar_staging/persisted_data"
}
```

I just wish I understood how this worked before.

natefoo:
Potentially fixes #313 but I haven't actually fully tested it, just verified that it resolves the exception. And it's not my favorite. @jmchilton?

### DIFF
--- a/pulsar/main.py
+++ b/pulsar/main.py
@@ -334,6 +334,16 @@ class PulsarManagerConfigBuilder(PulsarConfigBuilder):
         PulsarConfigBuilder.populate_options(arg_parser)
         arg_parser.add_argument("--manager", default=DEFAULT_MANAGER)
 
+    def load(self):
+        app_conf = super().load()
+        if self.manager != DEFAULT_MANAGER and "manager" in app_conf and "name" not in app_conf["manager"]:
+            log.info(
+                "Default manager not in app conf is not named but `--manager` is supplied, applying supplied manager "
+                f"name '{self.manager}' to default manager"
+            )
+            app_conf["manager"]["name"] = self.manager
+        return app_conf
+
 
 def main(argv=None, config_env=False):
     mod_docstring = sys.modules[__name__].__doc__
--- a/pulsar/manager_factory.py
+++ b/pulsar/manager_factory.py
@@ -32,7 +32,8 @@ def build_managers(app, conf):
             manager_description = ManagerDescription.from_dict(manager_options, manager_name)
             manager_descriptions.add(manager_description)
     elif "manager" in conf:
-        manager_description = ManagerDescription.from_dict(conf["manager"])
+        manager_name = conf["manager"].pop("name", DEFAULT_MANAGER_NAME)
+        manager_description = ManagerDescription.from_dict(conf["manager"], manager_name=manager_name)
         manager_descriptions.add(manager_description)
     else:
         manager_descriptions.add(ManagerDescription())


jmchilton:
None of the tests override this name. Is there a reason you cannot just use the default manage name? Oh... is it a message queue thing?

pulsar/main.py
```diff
@@ -334,6 +334,16 @@ def populate_options(cls, arg_parser):
...
 
+    def load(self):
+        app_conf = super().load()
+        if self.manager != DEFAULT_MANAGER and "manager" in app_conf and "name" not in app_conf["manager"]:
+            log.info(
+                "Default manager not in app conf is not named but `--manager` is supplied, applying supplied manager "
+                f"name '{self.manager}' to default manager"
+            )
+            app_conf["manager"]["name"] = self.manager
+        return app_conf
```
jmchilton, line 345: In puslar.py in the Galaxy job runner we have this in queue_job:

```
            remote_pulsar_app_config = dest_params.get("pulsar_app_config", {}).copy()
            if "pulsar_app_config_path" in dest_params:
                pulsar_app_config_path = dest_params["pulsar_app_config_path"]
                with open(pulsar_app_config_path) as fh:
                    remote_pulsar_app_config.update(yaml.safe_load(fh))
```

I feel like that might be the better place to put this? Or a better place.

jmchilton:
I feel like the whole approach might be mistaken though... the problem is different destinations are using the same amqp queue names? I kind of the think of the manager in the coexecution scenario as named _default_ and as the only manager - it isn't immediately clear why anyone would want to change that. Would the better approach here be to just key those on a different parameter?

pulsar/main.py
jmchilton: It looks like this block wouldn't be needed if you have a destination parameter declaring which manager to use - ```manager: tacc_k8s```. We've got code that is dispatching on the manager name and adding arguments at least:

```
    def _pulsar_script_args(self, manager_name, base64_job, base64_app_conf, wait_arg=None):
        manager_args = []
        if manager_name != "_default_":
            manager_args.append("--manager")
            manager_args.append(manager_name)
        if wait_arg:
            manager_args.append(wait_arg)
        manager_args.extend(["--base64", base64_job, "--app_conf_base64", base64_app_conf])
        return manager_args
```

The other block here is definitely good and I'll include a version of it in my PR.

mvdbeek:
I think this was superseded by https://github.com/galaxyproject/pulsar/issues/313#issuecomment-1424686260

PR #320, opened 2023-04-10
token endpoint

SergeyYakubov:
Add token endpoint parameter to client so that we can send it from Galaxy to Pulsar

See corresponding PR (https://github.com/galaxyproject/galaxy/pull/15300) and issue (https://github.com/galaxyproject/galaxy/issues/15526) for Galaxy


mvdbeek:
Merged as part of https://github.com/galaxyproject/pulsar/pull/321

PR #321, opened 2023-04-10
User auth plugins

SergeyYakubov:
Allows to configure plugins for user authentication/authorization. Includes methods to authenticate based on OIDC token (see issue https://github.com/galaxyproject/galaxy/issues/15526).

This is a stacked PR after https://github.com/galaxyproject/pulsar/pull/320

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,3 +32,4 @@ dependencies
 dependency_resolvers_conf.xml
 job_metrics_conf.xml
 .DS_Store
+.idea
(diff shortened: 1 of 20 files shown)


mvdbeek:
This looks good to me, besides the conflict that needs resolving I think my only comment is that it would be nice to have a small example of a configuration in the docs and how to verify it works.

mvdbeek:
I think you need to add pulsar.user_auth to https://github.com/galaxyproject/pulsar/blob/master/setup.py#L57 (or use `setuptools.find_packages` instead of the old manual way).

mvdbeek:
Thanks so much @SergeyYakubov, I'm sure we'll get the corresponding galaxy part in before the next release

PR #332, opened 2023-07-21
Add .readthedocs.yaml

natefoo:
This is required as of September. Stolen from bioblend, thanks @nsoranzo :grin:


natefoo:
I know I didn't cause it but I legitimately cannot find what the actual error is in the test output.

nsoranzo:
It has been failing for at least 5 months, so...

mvdbeek:
It's a test output that differs, but it's a pain to debug.

PR #333, opened 2023-07-27
Drop nose

neoformit:
Have replaced `nosetests` with `pytest` - may also want to remove entries from mypy.ini and setup.cfg.
Nose is EOL and doesn't work on Python >= 3.10.


mvdbeek (approved):
Thanks @neoformit!

neoformit:
Thanks @nuwang :laughing:

PR #339, opened 2023-09-21
Fail jobs when output files exist but stage out fails

natefoo:
With #258 we no longer fail jobs when work dir outputs cannot be staged out under any condition. This is good for getting back stdout/stderr when the tool fails for valid reasons and fails to produce workdir output. However, it also masks stage out failures, meaning that successful jobs that fail to stage out data will show up as green/ok in Galaxy but with zero length outputs.

This change still allows the job to succeed if workdir outputs are missing (more correctly, error handling still falls to Galaxy), but if staging out fails due to transport errors, then the job will be failed.

In my testing you still get back both job and tool stdout and stderr.

Draft because I'd like to have a custom client message on the Galaxy side for this if possible. Right now you get our old friend "Remote job server indicated a problem running or monitoring this job."

### DIFF
--- a/pulsar/client/exceptions.py
+++ b/pulsar/client/exceptions.py
@@ -6,11 +6,13 @@ Pulsar client exceptions
 class PulsarClientTransportError(Exception):
     TIMEOUT = 'timeout'
     CONNECTION_REFUSED = 'connection_refused'
+    NOT_200 = 'not_200'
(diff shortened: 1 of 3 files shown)


jmchilton:
Picking this line of work back up on an updated Pulsar in https://github.com/galaxyproject/pulsar/pull/505. Thanks for the contribution!

PR #357, opened 2024-03-26
Support configuring the AMQP heartbeat interval, or disabling heartbeats

natefoo:
@mvdbeek noticed in https://github.com/celery/celery/issues/1194#issuecomment-14566201 that heartbeats were considered unnecessary and disabled by default back in 2013. I am not sure about this (keeping connections alive across firewalls and so forth is sometimes necessary) but this adds the ability to disable the heartbeat and adjust the interval if so desired.

### DIFF
--- a/app.yml.sample
+++ b/app.yml.sample
@@ -93,6 +93,10 @@
 ## this value is used as the timeout argument to the producer.publish function.
 #amqp_publish_timeout: 2.0
 
+# AMQP heartbeat interval (see: https://www.rabbitmq.com/docs/heartbeats). Set
+# to false or 0 to disable heartbeats.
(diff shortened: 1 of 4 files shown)


jmchilton (approved):

jmchilton:
🤖 Posted by Claude (AI assistant) on John's behalf.

Superseded by #502, which rebases this commit onto current `master` with your authorship preserved, fixes two defects found while reviewing the rebase (a mutable default argument in `consume()` leaking a heartbeat between exchanges, and an uncoerced value reaching kombu), and adds the tests and docs the option needs. Thanks for the original patch — the option was the right idea.

Please reopen if this is wrong.

PR #365, opened 2024-04-26
Do not attempt to complete pre- or post-process if jobs are cancelled in the middle of either stage

### LINKED ISSUE #353: Do not retry on 403 error
When postprocessing, a 403 error typically means the job was cancelled by the user (except the strange case in #342). When this is the case, Pulsar keeps trying as per the retry settings, but the transfers will never succeed. It should probably just fail unconditionally in the event of a 403.

Alternatively, we could make configurable the error conditions to retry on, possibly with their own retry settings, for something like #298 where you might not want to retry as much, but maybe at least once or twice in case there's some filesystem caching weirdness.

### LINKED ISSUE #364: Outputs of deleted jobs are repeatedly uploaded to Galaxy until giving up when using `remote_transfer_tus`
Because Pulsar doesn't really have a great understanding of deleted jobs it will retry sending outputs and other postprocessing files back to Galaxy for deleted jobs. Under a normal remote_transfer action these are immediately rejected with a 403 and retried for however long the administrator has configured postprocessing retries (which it probably shouldn't do in the case of a 403, see #353). In the case of `remote_transfer` this is not great, but not terrible, just a little noisy and annoying.

In the case of `remote_transfer_tus`, the entire output is uploaded to Galaxy via tus, and _then_ the request is made to the job_files API that results in a 403. This results in a retry, at which point the entire output is uploaded to Galaxy _again_, followed by the 403. repeat and recycle up to the retry limit *for each and every output*.

We have `postprocess_action_max_retries` set to `10`. :skull: 

UPDATE: I spoke a bit too soon, it **does** record the deletion via [`manager._record_cancel()`](https://github.com/galaxyproject/pulsar/blob/9eb6a8e5232bd35b93882ec7c3c6133c502859ac/pulsar/managers/base/external.py#L28), [which causes postprocessing to be skipped](https://github.com/galaxyproject/pulsar/blob/9eb6a8e5232bd35b93882ec7c3c6133c502859ac/pulsar/managers/stateful.py#L215), it appears the condition arises if the job is deleted while postprocessing is taking place.

natefoo:
Mostly fixes #364. Ideally we would also interrupt any running transfers since those could be exceptionally long running and wasteful, but doing so requires interrupting a thread running a non-looping action.

We are probably better off moving those to more properly managed asynchronous tasks (celery, multiprocessing, etc.).

EDIT: Also fixes #353, although not by caring whether it's a 403.

### DIFF
--- a/pulsar/managers/staging/post.py
+++ b/pulsar/managers/staging/post.py
@@ -13,27 +13,27 @@ from pulsar.client.staging.down import ResultsCollector
 log = logging.getLogger(__name__)
 
 
-def postprocess(job_directory, action_executor):
+def postprocess(job_directory, action_executor, was_cancelled):
     # Returns True if outputs were collected.
     try:
         if job_directory.has_metadata("launch_config"):
             staging_config = job_directory.load_metadata("launch_config").get("remote_staging", None)
         else:
             staging_config = None
-        collected = __collect_outputs(job_directory, staging_config, action_executor)
+        collected = __collect_outputs(job_directory, staging_config, action_executor, was_cancelled)
         return collected
     finally:
         job_directory.write_file("postprocessed", "")
     return False
 
 
-def __collect_outputs(job_directory, staging_config, action_executor):
+def __collect_outputs(job_directory, staging_config, action_executor, was_cancelled):
     collected = True
     if "action_mapper" in staging_config:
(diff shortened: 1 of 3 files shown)


mvdbeek (approved):

mvdbeek:
Thanks @natefoo, should I cut a new release ?

PR #380, opened 2025-02-14
Prepare dirs from Galaxy, to properly recover resubmitted jobs

natefoo:
Galaxy side is in https://github.com/galaxyproject/galaxy/pull/16696


jmchilton:
We will see if there is a problem once the tests are fixed... ekkk.

PR #382, opened 2025-02-20
Add health check endpoint

nuwang:
Currently, there is no endpoint that can be queried to indicate a running pulsar. This makes it difficult for us to add liveness and readiness probes for the pulsar helm chart.


mvdbeek:
Thank you @nuwang!

PR #387, opened 2025-03-27
Add a deploy CI workflow

nsoranzo:
Will need [trusted publishing](https://docs.pypi.org/trusted-publishers/adding-a-publisher/) to be enabled on the 2 PyPI projects.


mvdbeek:
Worked like a charm, thank you @nsoranzo!

PR #399, opened 2025-06-30
Introduce file staging delegation via JSON staging manifest

kysrpex:
Introduce a JSON transfer action that indicates that the pulsar server should create a JSON manifest that can be used to stage files by an external system that can stage files in and out of the compute environment. This should be the general strategy for collecting input and output files for ARC, DIRAC, AWS batch etc.

Before launching jobs, Pulsar clients receive the staging manifest as a dictionary. When the job is complete, the script `pulsar-create-output-manifest` can be used to create a manifest declaring which files should be staged out by the external system.

This PR also updates the coexecutor Dockerfile so that it works again and so that it builds the Pulsar wheel (no need to build it separately anymore).

### DIFF
--- a/pulsar/scripts/collect_output_manifest.py
+++ b/pulsar/scripts/collect_output_manifest.py
@@ -1,3 +1,10 @@
+"""Write a JSON manifest containing the paths to the job outputs and the Galaxy URLs where they should be sent.
+
+This script is used by Pulsar clients that delegate staging out files to an external system, such as the ARC (Advanced
+Resource Connector) client. It is called after the job has finished to create a manifest declaring which files should be
+staged out by the external system.
+"""
+
 import argparse
 import json
 


kysrpex:
@mvdbeek That's all I needed to build the ARC integration (coming soon as a different PR). You may want to compare it to your branch [offline_connector_marius](https://github.com/mvdbeek/pulsar/tree/offline_collector_marius). I stripped out the Pulsar client (that comes in the other PR), the tests (please let me know which tests should make it into the PR) and a couple of things that were not needed to complete the integration.

kysrpex:
If this one is fine, then #370 should be closed.

docker/coexecutor/Dockerfile
```diff
@@ -1,41 +1,60 @@
-FROM conda/miniconda3
+FROM python:3.12-bookworm
```
kysrpex: @jmchilton [commented](https://github.com/galaxyproject/pulsar/pull/401#discussion_r2183060851):
> I don't think it is fair at all to say "Miniconda" is not required in the base image as the commit message suggests. I get that it isn't the modality that you wish to run it in but it is documented in https://pulsar.readthedocs.io/en/latest/containers.html#co-execution as an option and it is an option that makes a lot of sense to me. Is the slurm stuff required for your use case?

kysrpex: About the SLURM stuff, it is not needed for this use case but it was already in the old Dockerfile (it is needed for _other_ use cases).

```
    # Install packages
    && apt-get update \
    && apt-get install -y --no-install-recommends gcc \
        libcurl4-openssl-dev \
        cvmfs cvmfs-config-default \
        slurm-llnl slurm-drmaa-dev \
        bzip2 \
```

Because I am changing the base image to Debian Bookworm, the changes are required not to break the build with the new base image. `slurm-drmaa1` is not available on the Debian repos anymore for Bookworm (it has to be installed from the Galaxy Depot). The latest version of `slurm-drmaa1` available in the Galaxy depot (1.1.4) requires a library `libslurm36` from Debian Bullseye. There exists a [new version 1.1.5 of `libslurm-drmaa1` on GitHub](https://github.com/natefoo/slurm-drmaa/releases/tag/1.1.5) that maybe works on Bookworm without libraries from Bullseye, but @natefoo has not published a Debian package yet.

Issue #390, opened 2025-04-08
Implement GCP Batch Support


jmchilton:
Reviewing the docs @ https://cloud.google.com/batch/docs/reference/rest/v1alpha/projects.locations.jobs/create - it looks much closer to K8S & TES than AWS Batch (e.g. single document describing multiple containers) - so it should be easier to implement support for it here.

kysrpex:
I think the [ARC](https://www.nordugrid.org/arc/arc7/common/repos/repository.html) integration I am developing has at least some loose similarities. It would be nice if you ask me for a review when you open the PR (just to feel pressured to have a look and see how you did it, ofc you may ask more people to review). I'll do the same.

PR #404, opened 2025-07-11
Co-execution client for Google Cloud Platform Batch v1

jmchilton:
Implements #390.

Co-execution Diagram

From https://pulsar.readthedocs.io/en/latest/containers.html#galaxy-and-shared-file-systems about the advantages of a Pulsar co-execution client (a container native paradigm) over implementing a Galaxy job runner (an interface designed for shared file systems and DRMAA-style batch of jobs, not containers):

> The most glaring disadvantage of not using Pulsar in the above scenarios is that Galaxy must be deployed in the same container with the same mounts as the job execution environment. This prevents leveraging external cloud compute, multi-cloud compute, and makes it unsuitable for common Galaxy use cases such as large public instances, Galaxy’s leveraging institution non-cloud storage, etc… Even within the same cloud - a large shared file system can be an expensive prospect and Pulsar may allow making use of buckets and such more tractable.


pulsar/client/client.py
```diff
@@ -999,6 +1004,127 @@ def _raw_check_complete(self):
...
+        pass
+
+    def raw_check_complete(self) -> Dict[str, Any]:
+        gcp_job_params = self._gcp_job_params
+        job = get_gcp_job(gcp_job_params.project_id, gcp_job_params.region, self._tes_job_name, gcp_job_params.credentials_file)
+        status = job.status
+        state = status.state
+        return {
+            "status": gcp_state_to_pulsar_status(state),
+            "complete": "true" if gcp_state_is_complete(state) else "false",  # Ancient John, what were you thinking?
```
kysrpex, line 1075: For the ARC client implementation I am determining if the job is complete using the Pulsar state rather than the ARC job state.

```python
    def full_status(self):
        pulsar_state = self.get_status()
        return {
            "status": pulsar_state,
            "complete": "true" if manager_status.is_job_done(pulsar_state) else "false",
            # ancient John, what were you thinking? 👀
            "outputs_directory_contents": [],
            # it needs to be defined, otherwise `PulsarOutputs.has_outputs` fails; it is ok that it is empty because
            # ARC is responsible for staging the outputs (Galaxy does not have to collect any outputs)
        }
```

I get the  Pulsar state via a mapping too (similar to `gcp_state_to_pulsar_status()`). What would be the advantage of using two mappings `gcp_state_to_pulsar_status()`, `gcp_state_is_complete()` as you are doing for the GCP client implementation? (or what is my mistake?)


Would I need to use something akin to `gcp_state_is_complete()`

pulsar/client/client.py
```diff
@@ -999,6 +1004,127 @@ def _raw_check_complete(self):
...
+
+    def __init__(self, destination_params, job_id, client_manager):
+        super().__init__(destination_params, job_id, client_manager)
+
+
+class GcpPollingCoexecutionJobClient(BasePollingCoexecutionJobClient, LaunchesGcpContainersMixin):
+    """A client that co-executes pods via GCP and doesn't depend on amqp."""
+
+    def __init__(self, destination_params, job_id, client_manager):
+        super().__init__(destination_params, job_id, client_manager)
```
kysrpex, line 1059: The client manager is generated by Galaxy calling `pulsar.client.manager.build_client_manager()`. For the ARC client implementation I was forced to define a kwarg `arc_enabled` (and do some extra trickery). I assume this is not how things work. How are you managing to make Galaxy create a `PollingJobClientManager` rather than a `ClientManager` when you want to use the polling job client?

```
def build_client_manager(**kwargs: Dict[str, Any]) -> ClientManagerInterface:
    if 'job_manager' in kwargs:
        return ClientManager(**kwargs)  # TODO: Consider more separation here.
    elif kwargs.get('amqp_url', None):
        return MessageQueueClientManager(**kwargs)
    elif kwargs.get("k8s_enabled") or kwargs.get("tes_url") or kwargs.get("arc_enabled"):
        return PollingJobClientManager(**kwargs)
    else:
        return ClientManager(**kwargs)
```

kysrpex (commented):
I am not familiar enough with Pulsar to review this PR; but I can use this chance to ask questions and deliver a cleaner ARC client #401 implementation :)

pulsar/client/client.py
```diff
@@ -999,6 +1004,130 @@ def _raw_check_complete(self):
...
+    #                       used by the Job are still being cleaned up.
+    # CANCELLATION_IN_PROGRESS	The Job cancellation is in progress, this is because the resources used by the Job are
+    #                           still being cleaned up.
+    # CANCELLED The Job has been cancelled, the task executions were stopped and the resources were cleaned up.
+    state_map = {
+        batch_v1.JobStatus.State.STATE_UNSPECIFIED: manager_status.FAILED,
+        batch_v1.JobStatus.State.QUEUED: manager_status.PREPROCESSING,
+        batch_v1.JobStatus.State.SCHEDULED: manager_status.PREPROCESSING,
+        batch_v1.JobStatus.State.RUNNING: manager_status.RUNNING,
+        batch_v1.JobStatus.State.SCHEDULED: manager_status.COMPLETE,
```
ksuderman, line 1098: The key `batch_v1.JobStatus.State.SCHEDULED` appears twice in the `state_map` (and in the `state_map` in `gcp_state_is_complete` below.

jmchilton: Good catch - thank you!

pulsar/client/client.py
kysrpex: I did https://github.com/galaxyproject/galaxy/pull/20598/commits/6b6577527e941571707befe4c58f4ba337f51f8f, which is almost the same. However, I find your approach more attractive.

Could you mention me when you open a PR for this? I'd like to suggest something (merging `self.client_manager_kwargs` with the superclass' `client_manager_kwargs`) and adapt https://github.com/galaxyproject/galaxy/pull/20598 to your PR.

jmchilton: Will do. I'm sorry I have not been very responsive but I do appreciate your attention to the details.

PR #451, opened 2026-05-05
Allow a user mapping script for ExternalDrmaaQueueManager

bernt-matthias:
Use case: I'm currently trying to connect our HPC to usegalaxy.eu. The HPC requires submission as real user. Of course usernames on HPC and usegalaxy do not match. My idea is that I configure the destination on the Galaxy side to transfer  the email instead of the username to the pulsar server. On the pulsar server I can then use an LDAP lookup to get the username from the email.

### DIFF
--- a/docs/job_managers.rst
+++ b/docs/job_managers.rst
@@ -150,10 +150,17 @@ run DRMAA jobs via the user requested by the client (e.g. the Galaxy user).
         #chown_working_directory_script: scripts/chown_working_directory.bash
         #drmaa_kill_script: scripts/drmaa_kill.bash
         #drmaa_launch_script: scripts/drmaa_launch.bash
+        #user_mapping_script: 
 
 For more information on running jobs as the real user, check out `this discussion
 <http://dev.list.galaxyproject.org/Managing-Data-Locality-tp4662438.html>`__ from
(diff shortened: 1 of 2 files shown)


martincarrere:
@bernt-matthias a user can override his email after the creation of his account even when you have SSO has provider (I already reported that in Galaxy Project. For now my advice will be to keep this PR in draft until it's fixed if you want to do this in production.

bernt-matthias:
> @bernt-matthias a user can override his email after the creation of his account even when you have SSO has provider (I already reported that in Galaxy Project. For now my advice will be to keep this PR in draft until it's fixed if you want to do this in production.

True, instance should require account activation https://github.com/galaxyproject/galaxy/blob/d3b3ab7288cd272bb1d709f3c63b0f9ea440bc06/lib/galaxy/config/sample/galaxy.yml.sample#L1309 as usegalaxy.eu does. I tested this on eu and verified that users can not submit jobs as long as the new email is not verified.

martincarrere:
@bernt-matthias Ok so the combination of this other protection make it safe. Sorry for the warning i just wanted to be sure you dont open a gate. By the way I have the same exact issue to address to remap real users on my institute HPC. Thanks for this contribution.

PR #455, opened 2026-05-11
fix(docker): pin bookworm image and restrict whl glob to Python 3

martincarrere:
Fix the build of the docker image

- Pin both builder and final stages to `python:3.13-slim-bookworm` to prevent build failures
- Update whl glob to `*py3-none-any.whl` to explicitly target Python 3-only wheels, dropping Python 2 compatibility


martincarrere:
Hey @mvdbeek, if you have time, would appreciate a review. Cheers! 🦛

bgruening (approved):

PR #460, opened 2026-05-20
Fix queued_cli Slurm completion detection

dSizovs:
The parse_single_status method returned job_states.OK, which resolves to Job.states.OK = 'ok'. The stateful manager compares this against status.COMPLETE = 'complete', so the comparison always fails. FIX: Return 'complete' directly to match status.COMPLETE.

### DIFF
--- a/pulsar/managers/util/cli/job/slurm.py
+++ b/pulsar/managers/util/cli/job/slurm.py
@@ -62,7 +62,7 @@ class Slurm(BaseJobExec):
             id, state = status[1].split()
             return self._get_job_state(state)
         # else line like "slurm_load_jobs error: Invalid job id specified"
-        return "complete"
+        return job_states.OK
 
     def _get_job_state(self, state):
         try:
--- a/pulsar/managers/util/cli/job/slurm.py
+++ b/pulsar/managers/util/cli/job/slurm.py
@@ -62,7 +62,7 @@ class Slurm(BaseJobExec):
             id, state = status[1].split()
             return self._get_job_state(state)
         # else line like "slurm_load_jobs error: Invalid job id specified"
-        return job_states.OK
+        return "complete"
 
     def _get_job_state(self, state):
         try:
--- a/pulsar/managers/queued_cli.py
+++ b/pulsar/managers/queued_cli.py
@@ -5,15 +5,31 @@ qstat, etc...).
 
 from logging import getLogger
 
+from pulsar.managers import status
 from .base.external import ExternalBaseManager
 from .util.cli import (
     CliInterface,
     split_params,
 )
+from .util.cli.job import job_states
 from .util.external import parse_external_id
 from .util.job_script import job_script
 
 log = getLogger(__name__)
+
+# Translate the CLI plugins' job states into the Pulsar status vocabulary the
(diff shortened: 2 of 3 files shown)


bgruening:
@mvdbeek can you please look at this one?

mvdbeek:
Hmm, this doesn't seem quite right, other runners return OK, this is shared with Galaxy where OK is returned. I could be convinced with a test case though ? Maybe the manager needs an adapter there ?

pulsar/managers/util/cli/job/slurm.py
```diff
@@ -62,7 +62,7 @@ def parse_single_status(self, status, job_id):
             id, state = status[1].split()
             return self._get_job_state(state)
         # else line like "slurm_load_jobs error: Invalid job id specified"
-        return job_states.OK
+        return "complete"
```
mvdbeek: Ah! I think i get it, `job_states.OK` is complete when you don't have the galaxy model installed (i.e. it is assumed that we're on the pulsar server side), or else it is `"complete"` and we assume we're on the Galaxy side. This logic is broken though in multiple ways, we'd have to explicitly determine this instead of inferring it by what is importable here: https://github.com/galaxyproject/pulsar/blob/master/pulsar/managers/util/cli/job/__init__.py#L11

dSizovs: Exactly! My PR just patches the symptom in the Slurm plugin; I agree the real fix is determining the side explicitly rather than inferring it. Happy to leave that to you!

mvdbeek:
Thanks for working this out @dSizovs!

mvdbeek:
(and let us know how it goes from here!)

dSizovs:
Thank you @mvdbeek !
Will definitely update you!

dSizovs:
Thanks again @mvdbeek! Closing the loop since you asked. The fix works great in production: jobs from usegalaxy.eu now run on the NEMO HPC cluster via the relay and report status correctly e2e. I also upgraded to relay v0.2.x (Pulsar + the standalone pulsar-relay-client on Python 3.11) and it's all green. 
Thanks for the quick reviews, great first open-source experience.

pulsar/managers/util/cli/job/slurm.py
SergeyYakubov: Just hit the same problem. I'd argue that a real fix would be to have a single source of truth, e.g. always import relevant Galaxy modules in Pulsar and change Pulsar code accordingly.

dSizovs: Hi @SergeyYakubov !
Which Pulsar version are you on? @mvdbeek fixed this in 0.15.15. _CLI_STATE_TO_STATUS in queued_cli.py maps by enum member instead of value so it should work whether or not the Galaxy model is importable. Are you still seeing it on 0.15.15+?

SergeyYakubov: Hi @dSizovs,

We are using our local fork of 0.15.14. We cherry-picked this fix and it works now. My comment was more a thought about different logic for two cases - when Galaxy modules are importable or not - that happens in multiple places actually.

But we are fine for now, thank you both for the fix!

PR #466, opened 2026-06-19
Fix GCP Batch co-execution deadlocks

ksuderman:
Part of #465.

This PR ensures the tool and sidecar runnables are executed in the correct order and don't deadlock. 

This PR also:
1. Ensures outputs from intermediate jobs are collected.
Galaxy's path rewriting causes tools to write outputs into `working/` rather than `outputs/`, but `__collect_outputs()` only checked the outputs directory, so those files were skipped. Also relaxes the dynamic collection pattern from `dataset_\d+\.dat` to  `dataset_[\w-]+\.dat` so UUID-named datasets (e.g. `dataset_b0e57817-…​.dat`) match.
2. Re-raises infrastructure errors so they fail the job rather than producing a "successful" job with empty outputs. `FileNotFoundError` is deliberately excluded.
3. Pass `docker_extra_volumes` to the Batch containers
Comma-separated Docker `-v`-style volumes (used to bind-mount CVMFS paths such as
`/cvmfs/data.galaxyproject.org`) are now applied to both the sidecar and tool runnable
containers, so tools that depend on CVMFS reference data work.

Issue #465, opened 2026-06-19
Batch co-execution issue tracker


ksuderman:
This issue tracks the changes I made to benchmark the co-execution runner. I have grouped related changes to reduce the number of pull requests while also keeping each relatively easy to review and understand.

I will fill in the actual PR numbers once I open them.

### [PR 1](https://github.com/galaxyproject/pulsar/pull/466) — Co-execution deadlocks & output collection (critical)
The Pulsar co-execution runner uses two *runnables*: one that runs the tool itself, and a second *sidecar* that stages input data and collects results. Both were run as foreground runnables (the default) and GCP Batch runs foreground runnable in sequence waiting for the first to complete before starting the next.

This PR fixes the sidecar/tool runnable ordering so both containers run, collects outputs written to the `working/` directory (including UUID-named datasets), re-raises genuine infrastructure errors instead of swallowing them, and passes `docker_extra_volumes` (CVMFS) through to the Batch containers.

### [PR1.1](https://github.com/galaxyproject/pulsar/pull/467) — Re-raise infrastructure errors.
Split out from #466 and now just re-raises infrastructure errors (`ImportError`, `MemoryError`, `SystemError`, and infrastructure `OSError`s; disk full, I/O errors) so they fail the job .  `FileNotFound` and `403` errors still flow through.

### [PR 2](https://github.com/galaxyproject/pulsar/pull/472) — Custom VM boot disk image & size
Adds `custom_vm_image` and `boot_disk_size_gb` to `GcpJobParams` so jobs can boot
from a custom image (e.g. one with CVMFS pre-installed) and size the boot disk to fit it. Pairs with a [small companion change](https://github.com/galaxyproject/galaxy/pull/23068) in Galaxy's `PulsarGcpBatchJobRunner` (`PASSTHROUGH_PARAMS`) so the image URI can be set once at the runner level.

### [PR 3](https://github.com/galaxyproject/pulsar/pull/473) — Dynamic VM sizing & job lifecycle
Sizes the VM from TPV `cores`/`mem`, injects `GALAXY_SLOTS`/`GALAXY_MEMORY_MB` and sets the container compute resource so tools use the whole VM, validates local-SSD counts for N2/N2D machine types, generates collision-free job names with a configurable `job_id_prefix`, and makes job deletion on cancel optional (plus a `kill()` for the AMQP client) so jobs can be retained for debugging.

mvdbeek:
> critical

[...]

PR #467, opened 2026-06-19
Re-raise infrastructure errors during output collection

ksuderman:
Part of #465. 
Split out of #466

### What this does

`_collect_output` downgraded *all* collection exceptions to warnings for
`output_workdir` outputs (via `_allow_collect_failure`). That is correct for a missing
output, but it also swallowed genuine infrastructure failures — a disk-full or
out-of-memory during collection would surface as a "successful" job with empty outputs.

- Re-raise `ImportError`, `MemoryError`, `SystemError`, and infrastructure `OSError`s
  (disk full, I/O errors) so they fail the job.
- Two cases are deliberately **kept** recoverable and flow through the normal
  `_allow_collect_failure` handling:
  - `FileNotFoundError` — a missing output file is expected (e.g. a `from_work_dir`
    output a tool legitimately did not produce, which Galaxy represents as an empty
    dataset).
  - HTTP **403** — Galaxy authoritatively refused the upload (dataset purged mid-job);
    checked *before* the `OSError` test because `requests.HTTPError` is itself an
    `OSError` subclass.

Branch is based on current `master`.

### DIFF
--- a/pulsar/client/staging/down.py
+++ b/pulsar/client/staging/down.py
@@ -223,6 +223,9 @@ class ResultsCollector:
         log.info("collecting output {} with action {}".format(name, action))
         try:
             return self.output_collector.collect_output(self, output_type, action, name)
+        except (ImportError, MemoryError, SystemError):
+            # Genuine infrastructure failures must never be downgraded.
+            raise
         except Exception as e:
             if http_status_code(e) == 403:
                 # The Galaxy server authoritatively refused this upload (HTTP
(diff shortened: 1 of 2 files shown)


pulsar/client/staging/down.py
```diff
@@ -239,6 +245,14 @@ def _collect_output(self, output_type, action, name):
                     getattr(action, "url", None) or getattr(action, "path", action),
                 )
                 return False
+            if isinstance(e, OSError) and not isinstance(e, FileNotFoundError):
```
jmchilton: Can you move this logic into allow_collect_failure? It looks like there is an abstraction esbalished for this. You'll have to add the exception to the method interface. I think that is fine.

ksuderman: Is [this](https://github.com/galaxyproject/pulsar/pull/467/commits/1eb9aa1c232e0130a70867f3b2b1543f73237420) what you had in mind?

Issue #469, opened 2026-07-02
Resilience Suite flakiness at fixture startup


ksuderman:
The Resilience Suite job in the pulsar workflow intermittently fails at fixture setup with:
```
TimeoutError: Pulsar did not bind <mode> consumers within 60.0s
  test/resilience/harness/pulsar_control.py:199
```

The bind-wait helper in `harness/pulsar_control.py` documents this path as racy: "Polling the broker for a non-zero consumer count is racy after a kill+restart cycle: the old consumer count can linger." The 60s wait at pulsar_control.py:199 is the hard cutoff that trips.

PR #470, opened 2026-07-02
Fix resilience relay bind: drain relay poll-waiters on stop/kill

ksuderman:
Refs #469.

The `Resilience Suite` intermittently fails at **fixture setup** with
`TimeoutError: Pulsar did not bind relay consumers within 60.0s`
(`test/resilience/harness/pulsar_control.py`). 

## Cause

For AMQP modes, `PulsarControl.kill()` force-drops the stale broker consumer via
`_force_drop_setup_consumer_connections()` so the next setup can't race a lingering
registration. **Relay mode had no equivalent.** Pulsar's relay control consumer holds a
30 s `long_poll` waiter on the relay (`pulsar/messaging/bind_relay.py`); after a
stop/kill that waiter lingers on the relay until it notices the dropped connection.
Because the readiness check `_relay_has_pulsar_setup_waiter()` returns `True` on *any*
nonzero `*/job_setup` waiter, a following test's setup can pass/race against the dead
pulsar's registration.

## Fix

Deterministic relay teardown that mirrors the AMQP force-drop:

- Add `_relay_setup_waiter_count()` (returns `None` when the relay is unreachable, so
  "unknown" is distinguishable from a confirmed zero) and refactor
  `_relay_has_pulsar_setup_waiter()` to build on it.
- Add `_wait_relay_setup_waiters_drained()` — polls `/messages/poll/stats` until zero
  `*/job_setup` waiters; best-effort and bounded so it can never hang the suite.
- `stop()` and `kill()` await this drain in relay mode, so the following
  `wait_until_consuming()` only ever sees the freshly-started pulsar.

Harness-only change; no production code touched.

## Testing

- `py_compile` and `flake8` (max-line 150, complexity 14) pass.
- Not run against the live docker-compose stack — the failure is intermittent, so a
  single green run wouldn't prove the fix.

An independent interim mitigation (bind-timeout bump + fixture-setup retry) is proposed
separately in the `469-resilience-timeout` PR; the two are non-overlapping and can land
in either order.

### DIFF
--- a/pulsar/messaging/bind_relay.py
+++ b/pulsar/messaging/bind_relay.py
@@ -21,12 +21,18 @@ from .relay_state import RelayState
 
 log = logging.getLogger(__name__)
 
+DEFAULT_RELAY_LONG_POLL_TIMEOUT = 30.0
+
 
 def _server_cursor_path(manager) -> Optional[str]:
     persistence_directory = manager.persistence_directory
     if not persistence_directory:
         return None
     return os.path.join(persistence_directory, "%s-relay-cursor.json" % manager.name)
+
+
+def _relay_long_poll_timeout(conf):
+    return float(conf.get("relay_long_poll_timeout", DEFAULT_RELAY_LONG_POLL_TIMEOUT))
 
 
 def build_relay_transport(manager, relay_url, conf):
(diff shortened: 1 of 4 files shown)


mvdbeek:
Green on https://github.com/mvdbeek/pulsar/actions/runs/28662443620/job/85006053383, CI is severly hammered so I'll merge.

mvdbeek (approved):

ksuderman:
>CI is severly hammered so I'll merge.

That would be me opening a flurry of pull requests all at the same time 😁.  I am done now though and I'll stage bulk changes more slowly in the future.

PR #473, opened 2026-07-03
Dynamic VM sizing, resource env injection, and job lifecycle controls for GCP Batch

ksuderman:
# PR 3 — Dynamic VM sizing, resource env injection, and job lifecycle controls for GCP Batch

Part of #465.

## Summary

Makes the GCP Batch runner size VMs from per-tool resource requests, gives tools the
full VM resources, and adds operational controls for job naming and cleanup. Before
this PR every job used a fixed `machine_type`, tools ran single-threaded regardless of
the VM, job names could collide, N2/N2D local-SSD counts were rejected, and jobs were
always deleted on cancel (losing Cloud Logging for debugging).

## What's included

**1. Dynamic VM sizing from `cores`/`mem`.**
Adds `cores` and `mem` fields to `GcpJobParams`. When either is set,
`parse_gcp_job_params()` computes an appropriate N2 machine type via a new
`compute_machine_type()` helper in `gcp_util.py` (with `convert_cpu_to_milli()` and
`convert_memory_to_mib()`), choosing between `highcpu`/`standard`/`highmem` variants.
These helpers are intentionally Galaxy-free so Galaxy's direct
`GoogleCloudBatchJobRunner` can import them from Pulsar in a follow-up and share one
implementation.

**2. Inject `GALAXY_SLOTS` / `GALAXY_MEMORY_MB` and set the container compute resource.**
On a Batch VM there is no scheduler environment, so `$GALAXY_SLOTS` fell through to `1`
and tools ran single-threaded. When `cores`/`mem` are set, these env vars are injected
into the task environment and `task.compute_resource` is set so the container gets the
full VM instead of Batch's default 2 vCPU / 2 GB.

**3. Validate local-SSD count for N2/N2D machine types.**
Replaces a bare `assert size_gb % 375 == 0` with `_validate_ssd_size()`, which rounds up
to the nearest 375 GB and adds one more SSD for N2/N2D when the count is odd (these
types require an even count), logging the adjustment instead of crashing.

**4. Collision-free job names with a configurable prefix.**
Replaces the K8s-style name generator with `{prefix}-{job_id}-{unix_timestamp}` and
caches it so submit/poll/delete agree. Renames the internal helper
`gcp_galaxy_instance_id` → `gcp_job_id_prefix`, resolving the prefix as
`job_id_prefix` > `galaxy_instance_id` > `"pulsar"` so jobs from different Galaxy
instances sharing a project are distinguishable.

**5. Optional job deletion and AMQP `kill()`.**
Both client `kill()` paths now honor a `delete_batch_job` destination param (default
`true`); set `false` to retain jobs and their Cloud Logging for debugging. Adds the
previously missing `kill()` to `GcpMessageCoexecutionJobClient` and wraps deletion in
try/except so a failed delete doesn't cascade.

## Files changed

- `pulsar/client/client.py`
- `pulsar/client/container_job_config.py`
- `pulsar/managers/util/gcp_util.py`

## Testing

Verified VMs are sized from TPV `cores`/`mem`, tools run multi-threaded with the
injected `GALAXY_SLOTS`, N2 jobs provision with corrected SSD counts, job names no
longer collide, and jobs are retained when `delete_batch_job: false`. `flake8` clean.

## Notes

- Renames the internal `gcp_galaxy_instance_id` helper to `gcp_job_id_prefix`; if [PR 1](#466)
  has merged first this PR may need a trivial rebase. Recommended to merge last of the
  three.
- Future work (separate PR): update Galaxy's
  `galaxy/jobs/runners/util/gcp_batch/helpers.py` to re-export the sizing helpers from
  `pulsar.managers.util.gcp_util` so both runners share one implementation.

### DIFF
--- a/pulsar/client/client.py
+++ b/pulsar/client/client.py
@@ -1152,6 +1152,14 @@ class GcpMessageCoexecutionJobClient(BaseMessageCoexecutionJobClient, LaunchesGc
         super().__init__(destination_params, job_id, client_manager)
         self._setup_gcp_batch_client_properties(destination_params)
 
+    def kill(self):
+        if str(self.destination_params.get("delete_batch_job", "true")).lower() not in ("false", "0", "no"):
+            gcp_job_params = self._gcp_job_params
+            try:
+                delete_gcp_job(gcp_job_params.project_id, gcp_job_params.region, self._job_name, gcp_job_params.credentials_file)
+            except Exception:
+                log.warning("Failed to delete GCP Batch job %s", self._job_name)
+
 
 class GcpPollingCoexecutionJobClient(BasePollingCoexecutionJobClient, LaunchesGcpContainersMixin):
     """A client that co-executes pods via GCP and doesn't depend on amqp."""
@@ -1161,8 +1169,12 @@ class GcpPollingCoexecutionJobClient(BasePollingCoexecutionJobClient, LaunchesGc
         self._setup_gcp_batch_client_properties(destination_params)
 
     def kill(self):
-        gcp_job_params = self._gcp_job_params
-        delete_gcp_job(gcp_job_params.project_id, gcp_job_params.region, self._job_name, gcp_job_params.credentials_file)
+        if str(self.destination_params.get("delete_batch_job", "true")).lower() not in ("false", "0", "no"):
+            gcp_job_params = self._gcp_job_params
+            try:
+                delete_gcp_job(gcp_job_params.project_id, gcp_job_params.region, self._job_name, gcp_job_params.credentials_file)
+            except Exception:
+                log.warning("Failed to delete GCP Batch job %s", self._job_name)
 
     def clean(self):
         pass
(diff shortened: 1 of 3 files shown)


jmchilton:
I've been working on this review for a long time - I've gone down so many rabbit holes. I'm not sure the agents are aiding my understand or just weaponizing my ignorance 😭. I'm trying to fix up TES with changes like this one to bring them inline before merge instead of just letting diverge more and more. https://github.com/galaxyproject/pulsar/pull/491/changes. I'll PR something against your branch once I'm feeling more confident. Splitting point 4. Collision-free job names with a configurable prefix. out from the rest of this would really aid the review but isn't strictly needed.

jmchilton:
Can you rebase this on top of the now merged #493. I tried to preserve your commits. I think that will help me review it - sorry I've been so pedantic on these points.

ksuderman:
Yep. I will rebase and resolve conflicts once I am caught back up after some PTO.

PR #475, opened 2026-07-15
Native cvmfsexec support for CVMFS-hosted Singularity/Apptainer containers

natefoo:
Adds first-class support for running jobs under [cvmfsexec](https://github.com/cvmfs/cvmfsexec), so Singularity/Apptainer containers (and Galaxy reference data) hosted on CVMFS can be used on compute nodes that cannot mount `/cvmfs` system-wide, without root or special privileges. I've already been using this with great effect by hacking `env` `execute` statements in job destinations, but this is much better, plus it documents it and doesn't need a dirty hack for rewriting the container path when `mountrepo` mode is required.

## How it works

A `cvmfsexec` option on the job manager (`app.yml`), overridable per Galaxy destination:

```yaml
managers:
  _default_:
    type: queued_drmaa
    cvmfsexec:
      mode: mountrepo        # or "namespace"
      path: /path/to/cvmfsexec
      repositories:
        - data.galaxyproject.org
        - singularity.galaxyproject.org
```

Two modes:

- **namespace** (preferred): cvmfsexec provides a real `/cvmfs` inside a mount namespace and the job runs inside it. No path rewriting needed.
- **mountrepo**: for nodes without unprivileged user namespaces / fuse, repositories are bind-mounted under `<job_directory>/.cvmfsexec/dist/cvmfs`. The container image path (resolved by Galaxy against the real `/cvmfs`) is rewritten to the mounted location via an ordinary Galaxy `file_actions` `rewrite` rule.

The mount preamble is injected into the job script template (after dir prep); namespace mode wraps the command.

## Dependent Galaxy PR

mountrepo mode's image-path rewrite depends on a companion Galaxy change (https://github.com/galaxyproject/galaxy/pull/23129) that routes the resolved container image path through the compute environment's `unstructured_path_rewrite` (so a `file_actions` `rewrite` rule applies to it, the same way it applies to tool parameters).

## Docs & tests

- New Containers "Overview" and "Singularity / Apptainer and CVMFS" sections (cvmfsexec modes, building the self-contained distribution, the `galaxyproject.cvmfsexec` role, a usegalaxy.org example), plus the mountrepo rewrite example in the Galaxy Configuration docs.
- Unit and manager tests for config parsing, mount-preamble generation, namespace command wrapping, and the per-job `setup_params` override.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -9,7 +9,15 @@ History
 0.15.16.dev0
 ---------------------
 
-
+* Add native ``cvmfsexec`` support to Pulsar managers (``mountrepo`` and
+  ``namespace`` modes), configured via a ``cvmfsexec`` manager option or a
+  per-job override, for accessing CVMFS repositories (and CVMFS-hosted
+  container images) on hosts without a system-wide ``/cvmfs``.
+* Add a dedicated ``container`` file-action path type for rewriting resolved
+  container image paths (e.g. a Singularity/Apptainer image on CVMFS) via a
+  ``rewrite`` action, without overloading the ``unstructured`` tool-parameter
+  path type. **Note:** ``path_types: "*any*"`` now also matches container image
+  paths.
 
 ---------------------
 0.15.15 (2026-07-13)
--- a/app.yml.sample
+++ b/app.yml.sample
@@ -17,20 +17,19 @@
 #    #  ## real /cvmfs) or "mountrepo" (bind-mounts repos under
 #    #  ## <job_directory>/.cvmfsexec/dist/cvmfs for hosts without namespaces).
 #    #  mode: mountrepo
-#    #  ## Path to the cvmfsexec launcher to install into each job directory.
+#    #  ## Path to the cvmfsexec self-contained distribution to install into each
+#    #  ## job directory. See the "Containers" section of Pulsar's docs for how
+#    #  ## to build it: https://pulsar.readthedocs.io/en/latest/containers.html
 #    #  path: /path/to/cvmfsexec
 #    #  ## CVMFS repositories to mount for each job.
 #    #  repositories:
 #    #    - data.galaxyproject.org
 #    #    - singularity.galaxyproject.org
-#    ## In mountrepo mode the host-side container image path (resolved by Galaxy
-#    ## against the real /cvmfs) must be remapped to the mounted dist location.
-#    ## That is an ordinary path rewrite: configure it on the Galaxy job
-#    ## destination as a file_actions `rewrite` rule (source /cvmfs/<repo>,
-#    ## destination __PULSAR_JOB_DIRECTORY__/.cvmfsexec/dist/cvmfs/<repo>), not
-#    ## here. The __PULSAR_JOB_DIRECTORY__ token is substituted server-side with
-#    ## the absolute job directory (a $CVMFSEXEC_DIR shell var would be broken by
-#    ## the client's shlex.quote of the image path).
+#    ## In mountrepo mode the host-side container image path must be rewritten to
+#    ## the mounted location. This is necessary and is configured on the Galaxy
+#    ## job destination (not here) with a file_actions rewrite rule; see the
+#    ## example in the "Galaxy Configuration" section of Pulsar's docs:
+#    ## https://pulsar.readthedocs.io/en/latest/galaxy_conf.html
 
 
 ## Mode to create job releated directories with. If unset
--- a/app.yml.sample
+++ b/app.yml.sample
@@ -8,6 +8,26 @@
 #  _default_:
 #    type: queued_python
 #    num_concurrent_jobs: 1
+#    ## Run jobs under cvmfsexec (https://github.com/cvmfs/cvmfsexec) so CVMFS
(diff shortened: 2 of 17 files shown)


mira-miracoli:
That's a nice feature for @bernt-matthias I guess :)

jmchilton:
> **Posted by Claude (AI assistant) on behalf of @jmchilton — they did not write this text personally.**

Read through this alongside #23129 (commented there separately about the `pulsar-galaxy-lib` pin). The shape reads well — a dedicated `container` path type routed through the existing `FileActionMapper` rather than overloading `unstructured`, and a string-only `cvmfsexec` module with doctested config parsing. Three concrete notes and two softer ones.

### 1. `*any*` silently gains `container` — worth a `HISTORY.rst` entry

`ALL_PATH_TYPES` is read in exactly one place, the `*any*` expansion at `action_mapper.py:733`, so adding `CONTAINER` opts every existing `path_types: "*any*"` rule into container rewriting. The semantics seem right to me — "any path type" should mean any — but it changes behavior for deployments that never touched their `file_actions`. `docs/galaxy_conf.rst` documents the opt-in `path_types: container` case nicely; it's the `*any*` widening that's unannounced. The `0.15.16.dev0` section of `HISTORY.rst` is empty and this PR doesn't touch it.

Related, and more relevant once `*any*` matches: `PathMapper.check_for_container_rewrite` returns `None` when the matched action needs staging, so an existing `*any*` + `remote_transfer` rule now matches container paths and silently does nothing. The comment there already says "a staging action here would be a misconfiguration" — could that be a `log.warning` rather than a comment?

### 2. `namespace` mode doesn't carry the job script's shell environment across the wrap

`wrap_command` runs `cvmfsexec ... -- /bin/bash -c '<command>'`, and non-exported shell variables don't survive that. In `DEFAULT_JOB_FILE_TEMPLATE.sh`, `TMP`/`TEMP`/`TMPDIR` (`:39-51`), `GALAXY_PYTHON` (`:53`), `GALAXY_LIB` and `GALAXY_VIRTUAL_ENV` are all assigned without `export`. That's pre-existing and has never mattered, because nothing crossed a process boundary before this.

It's narrower than it first looks: by the time Pulsar sees the command line Galaxy has usually already containerized it, so the tool runs under `singularity exec` with its own env plumbing (`env_pass_through` / `singularity_env_*`). The return-code capture that `_expand_command_line` appends before the wrap is also fine — `$?` is evaluated by the inner shell and the rc path is absolute. What's exposed is whatever runs outside the container in the same command line.

`SingularityContainer.containerize_command` is the precedent for the fix: name the variables that have to cross rather than inheriting by accident. Given the module docstring already says namespace mode wants validation on a namespace-capable host, and the docs call it the preferred mode — worth either exporting those in the template, or saying plainly in `containers.rst` that namespace mode is provisional until it's been exercised?

### 3. Nothing in `cvmfsexec.py` is safely quoted except `command`

I first read this as "repository names are unquoted while paths are quoted," but the double quotes aren't doing the work either — `"%s"` still permits `$(...)`, backticks and `$VAR`:

```pycon
>>> 'cp "%s" "%s"' % ('/opt/$(id)', '/jobs/7/cvmfsexec')
'cp "/opt/$(id)" "/jobs/7/cvmfsexec"'
```

`shlex` is already imported. Switching the interpolations over is mechanical, and since `shlex.quote` leaves shell-safe strings alone it makes the common output *simpler*:

```python
commands = [
    "cp %s %s" % (shlex.quote(config.path), shlex.quote(launcher)),
    "%s -v >/dev/null" % shlex.quote(launcher),
    "trap %s EXIT" % shlex.quote("%s/umountrepo -a" % dist),
]
for repository in config.repositories:
    commands.append("%s %s" % (shlex.quote("%s/mountrepo" % dist), shlex.quote(repository)))
```

```pycon
>>> setup_commands(c, "/jobs/7")
['cp /opt/cvmfsexec /jobs/7/cvmfsexec',
 '/jobs/7/cvmfsexec -v >/dev/null',
 "trap '/jobs/7/.cvmfsexec/umountrepo -a' EXIT",
 '/jobs/7/.cvmfsexec/mountrepo singularity.galaxyproject.org']
```

plus `" ".join(shlex.quote(r) for r in config.repositories)` in `wrap_command`. This isn't a new privilege boundary — destination `env`/`execute` already hands Pulsar arbitrary shell through `env_to_statement` — but the per-job override arrives over the wire, so it seems worth closing in a brand-new file.

### 4 and 5 — things I noticed, not things I'm confident about

Entirely your judgement on both:

- The mount preamble has no failure handling. If `mountrepo` fails the script keeps going and the job dies later on a missing image path, which is unpleasant to diagnose from Galaxy's stderr. You'd know whether cvmfsexec's exit codes make a `||` guard worth having.
- `trap "... umountrepo -a" EXIT` takes the single global EXIT slot unconditionally. Nothing in Pulsar's template sets one today, so this is fine as it stands — mentioning it only because a destination `env: execute` entry could.

jmchilton:
@natefoo - anything more to do on this side? Do you need a release for the Galaxy side - can you handle that or does it require a me?

natefoo:
I think this is it for this side. I should be able to handle a release, theoretically, lol.

PR #482, opened 2026-07-31
Fix DRMAA FAILED job state reported as COMPLETE

gkr0110:
## Summary

`BaseDrmaaManager._get_status_external()` maps `JobState.FAILED` to
`status.COMPLETE`, with a comment already acknowledging this is wrong:

```python
JobState.FAILED: status.COMPLETE,  # Should be a FAILED state here as well
```

Any job that fails at the DRM submission or execution level (e.g. rejected
by Slurm for a missing/invalid account, or any other DRM-side failure) is
reported back to the Galaxy client as a successful completion, with no
output and no indication anything went wrong. We hit this concretely: jobs
submitted via `queued_drmaa` to Slurm through `slurm-drmaa` that got
`CANCELLED` for lacking `--account` were reported to Galaxy as `complete`
with an empty output directory.

`status.FAILED` already exists and `status.is_job_done()` treats `FAILED`
and `COMPLETE` identically for job-done detection, so this change only
affects the terminal status surfaced to the client, not job-done detection.

## Test plan

- [x] Confirmed via `git blame` this mapping (and stale comment) has been
      unchanged since it was introduced.
- [x] Grepped `test/` for references to `_get_status_external` /
      `JobState.FAILED` -- none found, so no existing tests assume the old
      (incorrect) behavior.
- [x] Reproduced against a real Slurm cluster via `slurm-drmaa`: a job
      submitted without an account is `CANCELLED` by Slurm and previously
      surfaced to the client as `complete`; with this change it correctly
      surfaces as `failed`.

### DIFF
--- a/pulsar/managers/base/base_drmaa.py
+++ b/pulsar/managers/base/base_drmaa.py
@@ -46,7 +46,7 @@ class BaseDrmaaManager(ExternalBaseManager):
             JobState.SYSTEM_SUSPENDED: status.QUEUED,
             JobState.USER_SUSPENDED: status.QUEUED,
             JobState.DONE: status.COMPLETE,
-            JobState.FAILED: status.COMPLETE,  # Should be a FAILED state here as well
+            JobState.FAILED: status.FAILED,
         }[drmaa_state]
 
     def _build_template_attributes(self, job_id, command_line, dependencies_description=None, env=[], submit_params={}, setup_params=None):


jmchilton:
Amazing - thank you for such a complete description of the issue, testing, etc.. and thanks for the fix! My agents found some polish they wanted to do at the higher layers and other managers based on this fix. Merging that PR should merge this one too - thanks again so much - great work! -John

gkr0110:
@jmchilton - It’s a tiny tweak, but I think it could be helpful for others. I spent a couple of hours trying to figure out what was breaking.. 

We are building a new Galaxy instance for Switzerland, so testing out the Pulsar workloads heavily. So expect more PRs/Issues 😄

jmchilton:
This fixed was included in 485 - thank you!

PR #489, opened 2026-08-19
Check import order in CI and sort the files that had drifted

jmchilton:
*This PR description was drafted and posted by Claude (AI assistant) on jmchilton's behalf.*

## Summary

`.isort.cfg` and the `make format` target have configured isort in this repo all along — the settings are nearly identical to Galaxy's (`profile=black`, `force_alphabetical_sort_within_sections`, `reverse_relative`, `force_grid_wrap=2`; only `line_length` and the skip paths differ). Nothing ever ran it in CI, so **71 files had drifted out of order**.

Two commits, and the second one is the boring one:

- **`Check import order in CI`** — the part worth reviewing. A `format` tox env running `isort --check --diff .`, plus a CI leg.
- **`Sort imports with isort`** — a mechanical `isort .` pass. No behavior change.

This came up because `lib/galaxy/jobs/runners/htcondor.py` failed Galaxy's isort check on galaxyproject/galaxy#23326, in an import block copied over from `pulsar/managers/queued_htcondor.py`. Galaxy caught it; Pulsar could not.

## Notes for review

**Why the format env runs on 3.14 only.** The lint job matrix covers 3.7 and 3.14. isort dropped 3.7 support after 5.11, so the two legs would resolve different isort versions and could disagree about the same file. A single 3.14 leg avoids that; this also mirrors how Galaxy keeps `format` separate from `lint`.

**Why it checks `.` rather than `pulsar test`.** That matches what `make format` rewrites, so `tools/` and `docs/` stay sorted too and the check never disagrees with the fixer.

**Dead config removed.** `setup.cfg` carried:

```
import-order-style = smarkets
application-import-names = pulsar,harness,recorder
```

Those configure `flake8-import-order`, which is not in the lint deps, so they have never had any effect. They also describe a *different* convention from the isort profile actually in use — smarkets wants `import X` before `from X import`, which black-profile isort does not enforce. If anyone installed the plugin the two would fight, so they are better gone than latent.

## Blast radius

`isort --check .` inspects the whole tree, so once this lands, open PRs branched from before it will fail the new check until they rebase. Worth timing accordingly.

## Validation

- `tox -e format` passes on the sorted tree; deliberately unsorting one import block makes it exit 1, so the check demonstrably catches drift rather than just passing.
- `tox -e test-unit` equivalent: 241 passed, 1 failed — `wsgi_app_test.py::test_standard_requests`, which fails identically on unmodified master.
- flake8 and mypy clean (188 source files).


test/user_authorization_test.py
```diff
@@ -1,7 +1,9 @@
+from test.test_utils import (
```
nsoranzo: This is getting confused with https://docs.python.org/3/library/test.html from the Python standard library. I'd suggest to use `from .test_utils import (` so it gets put back after `from unittest import TestCase`

test/resilience/scenarios/test_broker_outage.py
```diff
@@ -8,7 +8,6 @@
 
 import pytest
 import requests
-
```
nsoranzo, line 11: This looks wrong, harness is not a third-party library AFAIU.

nsoranzo (approved):
Thanks!

PR #497, opened 2026-09-10
Clarify real-user job configuration documentation

jmchilton:
Follow-up to #424.

This makes the new real-user guidance explicit about where `submit_user` is set and cleans up the grammar and punctuation.

Validation:
- `tox -e docs`
- `git diff --check`


bgruening (approved):

bgruening:
Thanks @jmchilton

PR #500, opened 2026-09-10
Close leaked file and transport handles

jmchilton:
Replacement for #367.

This carries the still-relevant parts of Marius van den Beek’s original cleanup onto current master, while omitting deleted modules, independently fixed code, and rewrites of paths that already closed handles safely.

It also closes resources missed by the original patch:

- pycurl `Curl` handles on success and error
- streaming Requests responses
- upload streams after exactly the blocking transfer lifetime

Marius retains authorship on the adapted original commit. The transport hardening and regression tests are isolated in a second commit.

Validation:

- `tox -e test-unit`: 315 passed, 63 skipped
- `tox -e mypy`: clean across 190 source files
- `tox -e lint,format`
- `git diff --check`

### DIFF
--- a/pulsar/client/action_mapper.py
+++ b/pulsar/client/action_mapper.py
@@ -463,7 +463,8 @@ class RemoteCopyAction(BaseAction):
         return RemoteCopyAction(source=action_dict["source"])
 
     def write_to_path(self, path):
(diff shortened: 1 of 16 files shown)


bgruening (approved):
Nice cleanup and fixes.

PR #505, opened 2026-09-14
Fail the job when a transport error blocks work-dir stage out

jmchilton:
> Posted by Claude (AI assistant) on jmchilton's behalf — they have reviewed this, but did not author the text.

Follow-up to #467. Its guard doesn't reach the transports that most deployments actually stage with, so the failure mode it set out to fix is still live.

### The gap

#467 stopped output collection from downgrading infrastructure failures to warnings, expressing "infrastructure failure" as `isinstance(exception, OSError)`. The curl and urllib transports raise `PulsarClientTransportError`, which subclasses plain `Exception` — a deliberate choice in #444, whose description notes it "subclasses `Exception`, so `except Exception` callers are unaffected." That compatibility choice is what slips past #467's guard five months later.

```
PulsarClientTransportError MRO: ['PulsarClientTransportError', 'Exception', 'BaseException', 'object']
is OSError subclass? False
pycurl.error               MRO: ['error', 'Exception', 'BaseException', 'object']
is OSError? False
```

Feeding each failure kind through `_allow_collect_failure` for an `output_workdir` output on current master:

| failure | downgraded? | |
|---|---|---|
| curl transport error | **yes** | wrong — job goes green with zero-length outputs |
| `requests.ConnectionError` | no | correct |
| disk full (`OSError(28)`) | no | correct |
| missing `from_work_dir` output | yes | correct, intended |

So it splits by transport. The **requests** transport is fine — `raise_for_status()` throws `requests.HTTPError`, which *is* an `OSError`. The **curl** transport is not.

Which one you get is not obvious: `pulsar/client/transport/__init__.py` binds the module-level `get_file`/`post_file` that `RemoteTransferAction` uses to curl whenever pycurl is importable, independently of `PULSAR_CURL_TRANSPORT` (which only steers `get_transport()`). pycurl is an optional, undeclared dependency, so whether a stage-out failure fails the job depends on whether an unrelated package is present in the environment.

This is the symptom natefoo reported in #339 in 2023: *"successful jobs that fail to stage out data will show up as green/ok in Galaxy but with zero length outputs."*

### Changes

**1. Convert pycurl errors in the staging transfers.** `PycurlTransport.execute` already wrapped `c.perform()` and translated `pycurl.error` into a structured `PulsarClientTransportError`, but `post_file` and `get_file` — the two functions staging actually uses — did not, so connection-level failures escaped as raw `pycurl.error`. Extracted that conversion into `_perform` and used it at all three call sites.

Also in this commit: a missing local file now raises `FileNotFoundError` rather than a bare `Exception` (staging policy keys on `FileNotFoundError` to decide what stays recoverable, so the two aren't interchangeable, and it matches the builtin and the requests transport); and non-200 responses carry a `NOT_200` code, so operators stop reading *"Unknown transport error (transport code: 500, ...)"* where the code is right there in the message.

**2. Consult a transport-error predicate in the collect policy.** Added `is_transport_error` beside the other transport-exception classifiers in `transient.py` — which already knows these exception shapes for `http_status_code` and `is_transient_http_error` — and used it in `_allow_collect_failure`. The 403 carve-out keys on the status code rather than the exception class, so it keeps applying to transport errors carrying 403.

### Tests

Five new tests, each verified red before the fix:

- `test_collect_output_reraises_transport_error_for_workdir_output` — the core case. Before: logged *"Allowed failure in postprocessing... Unknown transport error (transport code: 500)"* and returned normally.
- `test_collect_output_tolerates_403_transport_error_for_workdir_output` — the 403 carve-out still applies across exception classes.
- `test_curl_post_file_converts_connection_error` / `test_curl_get_file_converts_connection_error` — before: raw `pycurl.error: (7, "Couldn't connect to server")` escaped.
- `test_post_file_missing_file_raises_file_not_found`.

Full unit suite: 316 passed, 4 failed — the same 4 `/usr/bin/cow` macOS environmental failures present on unmodified master (verified in a clean baseline worktree). flake8 and mypy clean.

### Notes

- #339 (natefoo, draft since 2023) diagnosed this and is the origin of the `exceptions.py` and `down.py` shapes here. Its curl hunk landed independently via #444; the rest did not. That PR can be closed once this lands.
- Not addressed here: Galaxy still surfaces this as "Remote job server indicated a problem running or monitoring this job." A better client-side message was natefoo's stated reason for leaving #339 in draft, and it remains open work on the Galaxy side.
- An alternative to the explicit predicate would be making `PulsarClientTransportError` subclass `OSError`, which it arguably is. That fixes every `isinstance(e, OSError)` site at once but changes semantics anywhere `except OSError` is used, so I kept the blast radius small here.


mvdbeek:
There's a conflict here but i do like the thorough case assessment

mvdbeek (approved):

Issue #29, opened 2014-10-11
Follow up on __PULSAR_JOBS_DIRECTORY__


jmchilton:
- [ ] Document in Galaxy.
- [ ] Replace it in config files and metadata files.
- [ ] Write tests.
- [ ] Now that this technique is in place - go a step further and allow something like __PULSAR_JOB_DIRECTORY__ (without the s) so Galaxy doesn't even need to assume that job directory is `<jobs_directory>/<id>`.

jmchilton:
We found some serious issues in this feature and evidence of multiple sites working around them - so I think it is safe to say it just never worked well so I've removed the feature in #515 - rather than expanding the documentation.

PR #512, opened 2026-09-16
Substitute job directory tokens in staged files; stop clobbering jobs_directory

jmchilton:
Draft. The Pulsar half of #29. **This does not by itself make `jobs_directory: __PULSAR_JOBS_DIRECTORY__` work** — see "Still broken after this" below.

### 1. Staged config and metadata files kept the token

The TODO next to the command-line substitution has been there since 2014:

```python
# TODO: Handle __PULSAR_JOB_DIRECTORY__ in config files, metadata files, etc...
```

Only `command_line` was ever substituted. Both other channels are real:

- **Config files.** `client/staging/up.py` rewrites config-file *contents* with remote paths (`JobInputs.rewrite_paths`) and then uploads the rewritten text (`__upload_rewritten_config_files`). In token mode those paths are `__PULSAR_JOBS_DIRECTORY__/<id>/configs/...`. `tool_script.sh` is in this set, so it is not a corner case.
- **Metadata files.** `up.py:312-315` ships the whole metadata directory byte-for-byte — `handle_transfer_directory` with no `contents=`, so unlike config files these are not even rewritten in memory. `metadata/params.json` carries `filename_override` values that are remote output paths (Galaxy's `metadata/__init__.py:206, 362-367`), and under extended metadata a `compute_version_path` too.

Both arrive on the compute node holding the literal token.

**Where the substitution runs matters.** Not in `manager_endpoint_util.submit_job` — with remote staging the files do not exist yet at that point, and remote staging is exactly what `docs/files/job_conf_sample_mq_rsync.yml` (the only documented user of the token) configures. It runs in `StatefulManagerProxy` after `preprocess()` and before `launch()`, which is the one place both staging modes have converged.

New `pulsar/managers/util/job_directory_tokens.py` holds the token constants and the substitution helpers. `manager_endpoint_util` now calls the string helper instead of open-coding two `str.replace`es, so there is one definition of what these tokens mean rather than four.

Implementation notes that are deliberate, not incidental:

- Files are scanned **as bytes** for the token before any decode, so a job with no tokens pays one read per staged file and no decode.
- A file that holds a token but is not UTF-8 is logged and skipped. Metadata directories carry dataset blobs beside their JSON; decoding those would be wrong.
- Rewrites are in place, so `tool_script.sh` keeps its executable bit. There is a test for that.

### 2. An admin-set `jobs_directory` was silently discarded

`BaseRemoteConfiguredJobClient.__init__` guarded on `"job_directory"` (singular) and then assigned `"jobs_directory"` (plural):

```python
if "job_directory" not in destination_params:      # never present
    ...
    destination_params["jobs_directory"] = default_staging_directory
```

No destination param named `job_directory` exists anywhere in Pulsar or Galaxy — `_set_job_directory` reads only the plural key — so the guard never fired and `default_staging_directory()` always won. A Kubernetes, TES or GCP Batch destination that set `jobs_directory` had it replaced with `/pulsar_staging/` (or `/mnt/disks/<ssd_name>`). MQ clients were unaffected: their `default_staging_directory` returns `None`.

Unrelated to #29 and easy to split out if you'd rather.

### Still broken after this

The documented recipe stays unusable until Galaxy is fixed. `lib/galaxy/jobs/runners/pulsar.py:602` does:

```python
remote_job_directory = os.path.abspath(os.path.join(remote_working_directory, os.path.pardir))
```

In token mode `remote_working_directory` is *relative* (`__PULSAR_JOBS_DIRECTORY__/<id>/working`), so `abspath` prefixes Galaxy's own cwd, yielding `/<galaxy_cwd>/__PULSAR_JOBS_DIRECTORY__/<id>`. The token substring survives, so Pulsar's substitution still fires and produces a path that was never created. It reaches every command line via `command_factory.py:138-139` → `default_exit_code_file`.

The fix is to read `remote_job_config["job_directory"]`, which `client/setup_handler.py:99` already puts there, rather than re-deriving it. Galaxy PR to follow.

That this has gone unnoticed is itself worth weighing: the recipe may have no current users, in which case deleting the plural token could beat finishing it. Raised as an open question on #29.

### Verification

- Red-to-green on both fixes. The staged-file test drives a real `StatefulManagerProxy` over a real `QueueManager` and a real staging directory — it fails on master with the token still in the file, passes here.
- 19 new tests across `test/job_directory_tokens_test.py` (new), `test/stateful_test.py`, `test/manager_endpoint_util_test.py`, `test/client_test.py`.
- Full unit suite: **44 failed, 313 passed, 26 skipped**, against **44 failed, 294 passed, 26 skipped** on `ed076f2` with the same interpreter — same failure set (the macOS `/usr/bin/cow` environmental ones), +19 passing.
- `ruff check .`, `isort --check .`, `mypy pulsar test` (188 files) clean.

### Not in this PR

- Galaxy's `abspath` fix, and Galaxy-side documentation of the token (checkbox 1 — `git grep "__PULSAR"` over Galaxy's `dev` returns nothing at all).
- The sweep covers `configs/` and `metadata/`. Whether it should cover the whole job directory is an open design question on #29; inputs, outputs and `working/` hold job data rather than client-computed paths.


test/client_test.py
```diff
@@ -266,3 +267,24 @@ def test_tes_poll_and_cancel_use_recorded_external_id():
...
+    """Stands in for the coexecution clients, which all define a default."""
+
+    def default_staging_directory(self, destination_params):
+        return "/pulsar_staging/"
+
+
+def test_staging_default_applies_when_jobs_directory_absent():
+    client = _StagingDefaultClient({}, "123", None)
+    # The doubled separator comes from CONTAINER_STAGING_DIRECTORY's trailing
+    # slash and is pre-existing; asserted so a future change to it is deliberate.
```
jmchilton, line 282: Is this another bug? Shouldn't we just fix it?

jmchilton:
🤖 *Posted by Claude (AI assistant) on John Chilton's behalf — not authored by them personally.*

Closing this in favour of removing the feature outright — see #515.

This PR implemented box 2 of #29 (substituting job-directory tokens in staged config and metadata file contents). Investigating it further turned up a second, independent break in `__PULSAR_JOBS_DIRECTORY__` that this PR would not have fixed:

Setting `jobs_directory` to the token makes every client-computed path relative. Galaxy re-derives the job and tool directories from the working directory with `os.path.abspath` ([`pulsar.py:603-604`](https://github.com/galaxyproject/galaxy/blob/dev/lib/galaxy/jobs/runners/pulsar.py#L603-L604)), which on a relative path prefixes Galaxy's own working directory. Substitution then lands inside that prefix, so a single command line carried a correct `/staging/<id>/working` beside `/galaxy/cwd//staging/<id>` for the job directory and `tool_files`. `tool_script.sh` lives in the job directory, so little could run.

Both halves would have had to be fixed for the feature to work at all, and no release ever shipped either. Of the four public deployments that adopted the recipe, two commented it out again, one is a demo repo, and one is a python2.7-era playbook carrying a local patch that hand-rolled the config-file rewrite the feature never did.

Two things from this branch are worth keeping and have been split out:

- **The coexecution `jobs_directory` fix** (`61a9ca4`) is unrelated to #29 — `BaseRemoteConfiguredJobClient.__init__` guarded on `"job_directory"` while assigning `"jobs_directory"`, so a Kubernetes, TES or GCP Batch destination that set `jobs_directory` had it silently replaced. Now in its own branch, PR to follow.
- **Config-file substitution for the per-job `__PULSAR_JOB_DIRECTORY__` token** is still worth doing, and is much more tractable than it was for the plural token — the per-job token is resolved once, server-side, after staging, with no client-side path derivation involved. Tracking separately rather than bundling it with a removal.

PR #514, opened 2026-09-21
Use the manager name Galaxy returns from compute-resource registration

dSizovs:
Galaxy mints its own manager name at registration and routes jobs to it but register_with_galaxy returned the relay sub, so pulsar-config wrote the wrong name into app.yml. Registration reported success and jobs stayed queued. Use the name from Galaxy's response and fall back to the sub if it has none.

The existing happy-path test mocked Galaxy returning the sub as the manager name so it could not tell the two apart. Added tests where they differ.


jmchilton (approved):
I pushed some small tweaks stemming from an agentic review of the PR - I hope that is okay. Thanks so much for the fixes!

